PR #16, opened 2026-04-08
2.A.1: WS route accepts connections

Okpala76:
- RED: Initialized the create server for test with a random port
- Tested the Fastify server on a WebSocket route("/ws") 
- GREEN: updated the prod code to have Fastify initialize the websocket
- REFACTOR: created the plugin to thin out index file

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -1,7 +1,8 @@
-import Fastify from 'fastify';
 import fastifyStatic from '@fastify/static';
-import { resolve, dirname } from 'node:path';
+import Fastify from 'fastify';
+import { dirname, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
+import { websocketRoute } from './plugins/websocketRoutes.ts';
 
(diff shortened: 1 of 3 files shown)


KevinOrfas (commented):
**1. Typo:** `'Websocket fastify intergration test'` should be "integration".

**2. `console.log(socket)` in `websocketRoutes.ts:8`:** This will dump the entire socket object on every connection. Remove it or replace with actual handler logic.

**3. Test cleanup:** If the WebSocket connection fails, `server.close()` never runs. Use `afterEach` like the existing describe block does, otherwise a failing test leaves the server hanging.

**4. Type import:** `import { FastifyInstance } from 'fastify'` should be `import type { FastifyInstance }` since it's only used as a type annotation.

**5. Unnecessary string cast:** `const port = String(...)` is not needed, numbers work fine in template literals.

Core change is solid. Clean plugin pattern, proper use of `@fastify/websocket`, and the test verifies a real connection. Sort the above and it's good to merge.

Iam-Ebuka (approved):

PR #20, opened 2026-04-10
EKO-234: Story 2.B Client connects via WebSocket on page load

Okpala76:
## Summary

- `ClientSocket` class with null/real pattern matching the server side wrappers
- Unit tests (5) covering connection state, message send/receive, and URL validation
- Integration tests (5) covering real WebSocket connection, settled race condition guard, and auth token verification
- Wired up in `client/main.ts` so the connection opens on page load
- Learning material for the settled guard pattern in `learning/retro-settled-race-condition.md`

## What changed

| File | What |
|---|---|
| `client/clientSocket.ts` | New `ClientSocket` class with `create()`, `createNull()`, `RealClientSocket`, `StubbedClientSocket` |
| `client/main.ts` | Imports ClientSocket, connects on page load |
| `tests/client/clientSocket.test.ts` | Null tests: connection state, send, receive, URL validation |
| `tests/client/clientSocket.integration.test.ts` | Real tests: connect, settled guard paths, auth rejection/acceptance |
| `tests/infrastructure/fileStorage.integration.test.ts` | Cross platform fix for test directory paths |
| `learning/retro-settled-race-condition.md` | Retro writeup on the settled guard and test boundaries |

## Test plan

- [ ] `npm test` passes (unit tests)
- [ ] `npm run test:integration` passes (integration tests)
- [ ] Open app in browser, check DevTools > Network > WS tab shows active connection
- [ ] Close tab, confirm server `clientCount` drops

Linear: EKO-234

### DIFF
--- a/client/clientSocket.ts
+++ b/client/clientSocket.ts
@@ -63,7 +63,7 @@ export class StubbedServer {
 
 export class StubbedClientSocket implements ClientSocketInterface {
   private _isConnected = false;
-  private emiiter = new EventEmitter();
+  private emitter = new EventEmitter();
 
   get isConnected(): boolean {
     return this._isConnected;
@@ -80,7 +80,7 @@ export class StubbedClientSocket implements ClientSocketInterface {
 
   send(message: ClientMessage): Promise<void> {
     // Implementation for sending message
-    this.emiiter.emit(EVENT_MESSAGES, message);
+    this.emitter.emit(EVENT_MESSAGES, message);
 
     return Promise.resolve();
   }
@@ -90,10 +90,10 @@ export class StubbedClientSocket implements ClientSocketInterface {
   }
 
   onMessage(message: ServerMessage): void {
-    this.emiiter.emit('message', message);
+    this.emitter.emit('message', message);
   }
 
   trackMessages(): OutputTracker {
-    return new OutputTracker(this.emiiter, EVENT_MESSAGES);
+    return new OutputTracker(this.emitter, EVENT_MESSAGES);
   }
 }
--- a/client/clientSocket.ts
+++ b/client/clientSocket.ts
@@ -63,12 +63,19 @@ class RealClientSocket implements ClientSocketInterface {
 
   connect(): Promise<void> {
     return new Promise((resolve, reject) => {
+      let settled = false;
       this.socket = new WebSocket(this.url);
       this.socket.onopen = () => {
-        resolve();
+        if (!settled) {
+          settled = true;
+          resolve();
+        }
       };
       this.socket.onerror = (err) => {
-        reject(new Error(err.message satisfies string));
+        if (!settled) {
+          settled = true;
+          reject(new Error(err.message satisfies string));
+        }
       };
     });
   }
--- a/client/clientSocket.ts
+++ b/client/clientSocket.ts
@@ -1,3 +1,4 @@
+import { WebSocket } from 'ws';
 import { EventEmitter, OutputTracker } from '../server/infrastructure/output_tracker.ts';
 import { ClientMessage, ServerMessage } from '../shared/protocol.ts';
 
@@ -18,9 +19,9 @@ export class ClientSocket {
     this.client = client;
   }
 
-  //   static create(url: string): ClientSocket {
-  //     return new ClientSocket(new RealClientSocket(url));
-  //   }
+  static create(url: string): ClientSocket {
+    return new ClientSocket(new RealClientSocket(url));
+  }
 
   static createNull(): ClientSocket {
     // Implementation for null socket
@@ -45,6 +46,39 @@ export class ClientSocket {
   }
   trackMessages(): OutputTracker {
     return this.client.trackMessages();
+  }
+}
+
+class RealClientSocket implements ClientSocketInterface {
+  private socket: WebSocket | null = null;
+  private url: string;
+
+  constructor(url: string) {
+    this.url = url;
+  }
+
+  get isConnected(): boolean {
+    return this.socket !== null && this.socket.readyState === WebSocket.OPEN;
+  }
+
+  connect(): Promise<void> {
+    return new Promise((resolve) => {
+      this.socket = new WebSocket(this.url);
+      this.socket.onopen = () => {
+        resolve();
+      };
+      // this.socket.onerror = (err) => reject(err);
+    });
+  }
+
+  close(): Promise<void> {
+    return Promise.resolve();
+  }
+  send(): Promise<void> {
+    return Promise.resolve();
+  }
+  trackMessages(): OutputTracker {
+    return new OutputTracker(new EventEmitter(), '');
   }
 }
 
--- a/client/clientSocket.ts
+++ b/client/clientSocket.ts
@@ -1,0 +1,55 @@
+interface ClientSocketInterface {
+  connect(): Promise<void>;
+  close(): Promise<void>;
+  send(message: unknown): void;
+  get isConnected(): boolean;
+}
+
+export class ClientSocket {
+  private client: ClientSocketInterface;
+
+  private constructor(client: ClientSocketInterface) {
+    this.client = client;
+  }
+
+  //   static create(url: string): ClientSocket {
+  //     return new ClientSocket(new RealClientSocket(url));
+  //   }
+
+  static createNull(): ClientSocket {
+    // Implementation for null socket
+    return new ClientSocket(new StubbedClientSocket());
+  }
+
+  async connect(): Promise<void> {
+    await this.client.connect();
+  }
+
+  async close(): Promise<void> {
+    await this.client.close();
+  }
+
+  get isConnected(): boolean {
+    return this.client.isConnected;
+  }
+}
+
+export class StubbedClientSocket implements ClientSocketInterface {
+  private _isConnected = false;
+
+  connect(): Promise<void> {
+    this._isConnected = true;
+    return Promise.resolve();
+  }
+
+  close(): Promise<void> {
+    this._isConnected = false;
+    return Promise.resolve();
+  }
+
+  send(): void {}
+
+  get isConnected(): boolean {
+    return this._isConnected;
+  }
+}
--- a/client/clientSocket.ts
+++ b/client/clientSocket.ts
@@ -1,4 +1,4 @@
-import { WebSocket } from 'ws';
+import { WebSocket, ClientOptions } from 'ws';
 import { EventEmitter, OutputTracker } from '../server/infrastructure/output_tracker.ts';
 import { ClientMessage, ServerMessage } from '../shared/protocol.ts';
 
@@ -19,12 +19,13 @@ export class ClientSocket {
     this.client = client;
   }
 
-  static create(url: string): ClientSocket {
+  static create(url: string, options: ClientOptions = {}): ClientSocket {
     const parsed = new URL(url);
     if (parsed.protocol !== 'ws:' && parsed.protocol !== 'wss:') {
       throw new Error(`Invalid WebSocket URL: expected ws:// or wss://, got ${parsed.protocol}`);
     }
-    return new ClientSocket(new RealClientSocket(url));
+
+    return new ClientSocket(new RealClientSocket(url, options));
   }
 
   static createNull(): ClientSocket {
@@ -55,10 +56,12 @@ export class ClientSocket {
 class RealClientSocket implements ClientSocketInterface {
   private socket: WebSocket | null = null;
   private readonly url: string;
+  private readonly options: ClientOptions;
   private emitter = new EventEmitter();
 
-  constructor(url: string) {
+  constructor(url: string, options: ClientOptions = {}) {
     this.url = url;
+    this.options = options;
   }
 
   get isConnected(): boolean {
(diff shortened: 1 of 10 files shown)


KevinOrfas (commented):
Nice work so far! The TDD rhythm in the commits is clean and the null pattern mirrors the server side well. A few things to look at when you pick this back up:

## Small fixes

- **Typo: `emiiter`** in `client/clientSocket.ts` line 68 — should be `emitter`
- **Typo: `recieve`** in the test at `tests/client/clientSocket.test.ts` line 25 — should be `receive`

## Design thoughts

- **`StubbedClientSocket` and `StubbedServer` are both exported** — on the server side, `StubbedWebSocketServer` is kept private and only `StubbedClient` is exported (because tests access it through `simulateConnection()`). Same idea here: consumers should go through `ClientSocket.createNull()` and `simulateServer()`, so `StubbedClientSocket` doesn't need to be exported directly.

- **Sent and received messages share the same tracker** — right now `trackMessages()` emits on `'message'` for both `send()` and `onMessage()`. That means if you send a client message and then receive a server message, `tracker.data` will have both mixed together. On the server side, connections, disconnections, and messages each have their own event and tracker. Worth thinking about whether the client should distinguish between outbound and inbound, otherwise the tracker won't be very useful beyond single-action tests.

## What's left (from the story)

I know the PR description says work in progress, so just listing these so they're visible:

- [x] `create()` factory method (currently commented out)
- [x] Wire it up in `main.ts` so it connects on page load
- [x] Integration test (`clientSocket.integration.test.ts`) with a real WebSocket server

No rush on any of this. Give me a shout if you want to pair on the integration test or the `main.ts` wiring.

KevinOrfas:
# Security First: What PR #20 Taught Us

> "Education must begin with the solution of the student teacher contradiction, by reconciling the poles of the contradiction so that both are simultaneously teachers and students." Paulo Freire

This is not a lecture. This is a conversation. Read the code, answer the questions, and talk to your pair about what you notice. The goal is not to memorise rules. The goal is to train your eye so you see these things before anyone has to point them out.

## 1. The Hanging Promise

Here is the original `connect()` method. Read it carefully.

```ts
connect(): Promise<void> {
  return new Promise((resolve) => {
    this.socket = new WebSocket(this.url);
    this.socket.onopen = () => {
      resolve();
    };
    // this.socket.onerror = (err) => reject(err);
  });
}
```

**Stop and think:**

- What happens if the server is not running?
- What happens if the URL is wrong?
- The promise resolves on `onopen`. What resolves it on failure?
- If nothing resolves or rejects a promise, what happens to the caller that `await`s it?

**Try it yourself.** Paste this into a scratch file and run it:

```ts
const never = new Promise((resolve) => {
  // imagine the callback never fires
});

console.log('before await');
await never;
console.log('after await'); // does this ever print?
```

**The lesson:** Every `new Promise` needs both a resolve path AND a reject path. If you only handle the happy path, the sad path hangs forever. No error, no crash, just silence. That is worse than a crash because nobody knows something went wrong.

**Now look at the fix:**

```ts
connect(): Promise<void> {
  return new Promise((resolve, reject) => {
    let settled = false;
    this.socket = new WebSocket(this.url, this.options);
    this.socket.onopen = () => {
      if (!settled) {
        settled = true;
        resolve();
      }
    };
    this.socket.onerror = (err) => {
      if (!settled) {
        settled = true;
        reject(new Error(err.message satisfies string));
      }
    };
  });
}
```

**Questions for the pair:**

- Why is `reject` wrapped in `new Error()`? What happens if you reject with something that is not an Error?
- What is the `settled` flag doing? Can both `onopen` and `onerror` fire? (Yes, they can. Google "WebSocket onerror onclose order".)
- If we removed the `settled` guard, what would happen when both events fire?

## 2. The URL You Never Checked

Original code:

```ts
static create(url: string): ClientSocket {
  return new ClientSocket(new RealClientSocket(url));
}
```

**Stop and think:**

- What happens if someone passes `http://localhost:8080`?
- What happens if someone passes `ftp://dodgy-server.com/payload`?
- The `ws` library (the Node.js WebSocket client we use) will try to connect to anything. Should your code trust that every string it receives is a valid WebSocket URL?

**The principle:** Validate at the boundary. When data enters your system from outside (user input, config, function parameters), check it before using it. Do not assume the caller got it right.

**The fix:**

```ts
static create(url: string, options: ClientOptions = {}): ClientSocket {
  const parsed = new URL(url);
  if (parsed.protocol !== 'ws:' && parsed.protocol !== 'wss:') {
    throw new Error(
      `Invalid WebSocket URL: expected ws:// or wss://, got ${parsed.protocol}`
    );
  }
  return new ClientSocket(new RealClientSocket(url, options));
}
```

**Questions for the pair:**

- What does `new URL(url)` do if the string is not a valid URL at all? (Try it in the console.)
- Why do we check protocol specifically? What attack could a bad protocol enable?
- Should this check live in `create()` or in `RealClientSocket`'s constructor? What is the tradeoff?

## 3. The Open Door

Here is the original integration test server:

```ts
webSocketServer = WebSocketWrapper.createRawWs({ port: PORT });
await webSocketServer.start();
const client = ClientSocket.create('ws://localhost:9877');
await client.connect();
// connected! no questions asked
```

**Stop and think:**

- The server accepted the connection. Did it ask who the client was?
- Could any process on the machine connect to this port?
- In production, could any browser on the internet connect?
- What is the difference between authentication (who are you?) and authorisation (are you allowed to do this?)?

**Now look at the auth test:**

```ts
function createAuthServer(): WebSocketServer {
  return new WebSocketServer({
    port: PORT,
    verifyClient: (info, cb) => {
      const token = info.req.headers['authorization'];
      if (!token) {
        cb(false, 401, 'Unauthorized');
        return;
      }
      cb(true);
    },
  });
}

it('rejects connections without an auth token', async () => {
  rawServer = createAuthServer();
  const client = ClientSocket.create(`ws://localhost:${String(PORT)}`, {});
  await expect(client.connect()).rejects.toThrow();
});

it('connects when a valid auth header is provided', async () => {
  rawServer = createAuthServer();
  const client = ClientSocket.create(`ws://localhost:${String(PORT)}`, {
    headers: { Authorization: 'Bearer test-token' },
  });
  await client.connect();
  expect(client.isConnected).toBe(true);
  await client.close();
});
```

**Questions for the pair:**

- The `verifyClient` callback runs during the HTTP upgrade handshake, before the WebSocket connection is established. Why is that important? What if we checked auth after the connection was open?
- This test accepts any non empty token. What would a real implementation need to check?
- Where should the auth check live: the client, the server, or both? Why?
- What is `Bearer` in `Bearer test-token`? Look up "RFC 6750".

## 4. The Tests That Lied

The original integration test had this:

```ts
webSocketServer
  .close()
  .then(() => {
    expect(client.isConnected).toBe(false);
  })
  .catch(() => {
    expect(client.isConnected).toBe(true);
  });
```

This test passed. But the assertions inside `.then()` and `.catch()` never actually ran.

**Stop and think:**

- The test function is `async`. The `.then()` chain is not `await`ed. What happens?
- The test runner sees no errors thrown synchronously, so it reports green. But did it actually test anything?
- How would you prove that a test assertion is actually running? Try it.

**The lesson:** A green test is not the same as a good test. A test that never runs its assertions is worse than no test at all because it gives you false confidence. Always `await` your promises in async tests.

## 5. The Cleanup That Wasn't

Original `afterEach`:

```ts
afterEach(async () => {
  await server.close(); // closes the Fastify server
});
```

But the test created a `webSocketServer` on port 9877 that was never cleaned up in `afterEach`.

**Stop and think:**

- What happens to a port when a server is not properly closed?
- If this test file had two tests, what would happen when the second test tries to start a server on the same port?
- What is a resource leak? Can you name three types? (File handles, database connections, network ports...)

**The principle:** If you open it, close it. If you start it, stop it. Always in `afterEach`, never inside the test body, because `afterEach` runs even when the test fails.

## Your Turn

Pick one of these patterns and find it in another part of the codebase. Does our server code validate URLs? Do we clean up every resource we open? Where else might a promise hang?

Look at the code you wrote last week. What would you change?

> "Knowledge emerges only through invention and reinvention, through the restless, impatient, continuing, hopeful inquiry human beings pursue in the world, with the world, and with each other." Paulo Freire

KevinOrfas:
# Closures: A Show and Tell

> At the dojo we understand what our tools do. We do not just use them. We know how they work.

## What is a closure?

A closure is when a function remembers variables from the place where it was defined, even after that place has finished executing.

That is it. That is the whole concept.

## Show me

Open `tests/client/clientSocket.integration.test.ts` and look at this:

```ts
describe('ClientSocket auth', () => {
  const PORT = 9878;

  function createAuthServer(): WebSocketServer {
    return new WebSocketServer({
      port: PORT,   // <-- where does PORT come from?
      verifyClient: (info, cb) => {
        const token = info.req.headers['authorization'];
        if (!token) {
          cb(false, 401, 'Unauthorized');
          return;
        }
        cb(true);
      },
    });
  }

  it('rejects connections without an auth token', async () => {
    rawServer = createAuthServer();
    const client = ClientSocket.create(`ws://localhost:${String(PORT)}`);
    // ...
  });
});
```

`createAuthServer` is defined inside the `describe` block. `PORT` is defined in the `describe` block. The function uses `PORT` even though `PORT` is not a parameter and not defined inside the function.

That is a closure. `createAuthServer` "closes over" the variable `PORT`.

## Why should I care?

Because closures are everywhere in JavaScript and you are already using them. Every callback you have ever written inside a function is probably a closure.

### Example 1: Event handlers

```ts
this.socket.onopen = () => {
  resolve();  // resolve comes from the enclosing Promise constructor
};
```

`resolve` is not defined inside the arrow function. It comes from the `new Promise((resolve, reject) => { ... })` that wraps it. The arrow function closes over `resolve`.

### Example 2: The settled guard

```ts
connect(): Promise<void> {
  return new Promise((resolve, reject) => {
    let settled = false;                    // defined here

    this.socket.onopen = () => {
      if (!settled) {                       // used here
        settled = true;                     // mutated here
        resolve();
      }
    };

    this.socket.onerror = (err) => {
      if (!settled) {                       // and here
        settled = true;                     // and here
        reject(new Error(err.message));
      }
    };
  });
}
```

Two separate functions (`onopen` handler and `onerror` handler) both close over the same `settled` variable. When one sets it to `true`, the other sees the change. They share the same variable, not a copy of it.

**Question:** What would happen if each function got its own copy of `settled` instead of sharing the same one? Would the race condition guard still work?

### Example 3: The OutputTracker

```ts
export class OutputTracker {
  private _data: unknown[] = [];

  constructor(emitter: EventEmitter, eventType: string) {
    emitter.on(eventType, (data: unknown) => {
      this._data.push(data);   // this arrow function closes over `this`
    });
  }
}
```

The callback passed to `emitter.on` will fire later, possibly much later, long after the constructor has returned. But it still has access to `this._data` because the arrow function closes over `this` from the constructor scope, and then accesses `_data` as a property on it. Arrow functions do not have their own `this`. They always use the one from where they were defined.

## The mental model

Think of it like a backpack. When a function is created, it packs up all the variables from its surrounding scope and carries them along. Wherever that function goes, whatever calls it, it still has its backpack.

```
┌─────────────────────────────────────┐
│ describe('ClientSocket auth')       │
│                                     │
│   const PORT = 9878                 │
│                                     │
│   function createAuthServer() {     │
│     ┌─────────────┐                │
│     │  backpack:   │                │
│     │  PORT = 9878 │                │
│     └─────────────┘                │
│     return new WebSocketServer({    │
│       port: PORT   // from backpack │
│     });                             │
│   }                                 │
│                                     │
└─────────────────────────────────────┘
```

## Common traps

### Trap 1: The loop variable

```ts
// broken
for (var i = 0; i < 3; i++) {
  setTimeout(() => console.log(i), 100);
}
// prints: 3, 3, 3 (not 0, 1, 2)
```

Why? All three arrow functions close over the same `i`. By the time they run, the loop is done and `i` is 3.

```ts
// fixed with let (block scoped, each iteration gets its own)
for (let i = 0; i < 3; i++) {
  setTimeout(() => console.log(i), 100);
}
// prints: 0, 1, 2
```

**Question:** Why does `let` fix this but `var` does not? What is the difference in scoping?

### Trap 2: Accidental memory leaks

If a closure holds a reference to a large object, that object cannot be garbage collected as long as the closure exists.

```ts
function setup() {
  const hugeArray = new Array(1_000_000).fill('data');

  return () => {
    console.log(hugeArray.length);  // hugeArray lives forever now
  };
}

const leakyFn = setup();
// setup() finished, but hugeArray is still in memory
// because leakyFn's closure holds a reference to it
```

**Question:** How would you fix this? When is it OK and when is it a problem?

## Try it yourself

1. Open a Node REPL (`node` in your terminal)
2. Read the code below and predict the output before running it. Write your predictions down. Then type it in:

```js
function makeCounter() {
  let count = 0;
  return {
    increment: () => ++count,
    getCount: () => count,
  };
}

const counter = makeCounter();
console.log(counter.getCount());   // what does this print?
counter.increment();
counter.increment();
console.log(counter.getCount());   // what about this?
```

3. Now make a second counter: `const counter2 = makeCounter()`. Does incrementing `counter2` affect `counter`? Why or why not?

## The one sentence version

A closure is a function that remembers the variables from where it was born.

Every time you write a callback, an event handler, or a function inside another function, you are using closures. Now you know what to call it.

> At EkoHacks we do not just write code that works. We understand why it works.

PR #23, opened 2026-04-16
Naming consistency cleanup

Iam-Ebuka:
# EKO-267: naming consistency across wrappers

Slack time cleanup. No feature change, no behaviour change. Names only.

## The pattern

Three small rules applied across every wrapper.

**Public wrapper classes end with `Wrapper`.** `FileStorage` becomes `FileStorageWrapper`. `ScriptRunner` becomes `ScriptRunnerWrapper`. `ClientSocket` becomes `ClientSocketWrapper`. This aligns with `MongoWrapper` and `WebSocketWrapper`, which already followed the pattern.

**Interfaces drop filler words.** `WebSocketServerInterface` becomes `WebSocketInterface`. `MongoClientInterface` becomes `MongoInterface`. `FileSystemInterface` becomes `FileStorageInterface`. `ProcessRunnerInterface` becomes `ScriptRunnerInterface`. Each file has one interface, so the extra nouns were not earning their keep.

**Internal classes align with the domain, not the library.** `RealWebSocketServer` becomes `RealWebSocket`. `FastifyWebSocketServer` becomes `FastifyWebSocket`. `StubbedWebSocketServer` becomes `StubbedWebSocket`. Same shape for `RealMongo`, `StubbedMongo`, `RealFileStorage`, `StubbedFileStorage`, `RealScriptRunner`.

Plus one file rename to match the project's camelCase convention: `server/infrastructure/output_tracker.ts` becomes `outputTracker.ts`.

## What to check in review

- All callers updated across `client/`, `server/`, and `tests/`. Worth a grep for the old names to be sure nothing was missed.
- No behaviour change. The test suite should be identical green before and after.
- The file rename will show up as a move in git. Anyone running `git log` on the old path will need the new one.

## Why now

Consistency compounds. Before we build the 3.A publications layer on top of these wrappers, naming them the same way makes the next engineer's load a bit lighter.


Iam-Ebuka:
https://linear.app/ekohacks/issue/EKO-267/ekoslack-naming-consistency-cleanup

Iam-Ebuka:
The reason for leaving existing and adding suffix of wrapper to facades is to maintain consistency. We cannot use remove wrapper suffix from WebsocketWrapper because the name 'websocket' will collide with the 'type websocket' imported from 'ws'.

KevinOrfas:
# EKO-267 review

CI is green. Nothing here blocks the merge: unit tests pass, typecheck is clean, lint's only complaint is the `console.log` in `start.ts` that has been there for weeks. Seventeen commits of disciplined, careful renaming, and the result is real: the codebase now has a convention that wasn't written down before. `Wrapper` suffix on every public injectable, `Real*` / `Stubbed*` / `Fastify*` on the internal backends, `*Interface` on the interfaces, camelCase filenames everywhere.

Three things stuck with me while reading. None block the merge, all worth a conversation, either now or as a follow up.

First, `RealWebSocket`. `server/infrastructure/websocket.ts` imports `type WebSocket` from the `ws` library at the top. Two lines down, we now have a class called `RealWebSocket`. A reader scanning the file sees `WebSocket` (the connection type from the library) and `RealWebSocket` (our server implementation) sitting together. The old name, `RealWebSocketServer`, was unambiguous. Is the shorter name worth the cost at the reading site, or does the ambiguity matter?

Second, `RealMongo` and `StubbedMongo`. The MongoDB driver itself calls its connection class `MongoClient`. The old `RealMongoClient` mirrored that. Once the `Client` suffix is gone, readers have to decide from context whether `Mongo` refers to the database, the driver, or the connection. Is that context cost acceptable, or do we want to keep the `Client` anchor?

Third, and this is the one I'd love your take on: the rules you are enforcing in this PR do not exist anywhere except in the commits. Reviewers of future rename work will have to do the same archaeological reading to find the convention you just set. Two honest ways to go. Write the rules down somewhere searchable and commit to maintaining them. Or keep conventions tacit and trust that whoever reviews the next PR will read the pattern off the codebase. Both have real costs. Which one fits how our team works?

Two smaller things, neither blocking:

- Every commit is tagged `fix:`. Conventional commits usually reserve `fix:` for bug fixes and `refactor:` for renames with no behaviour change. Which prefix do we want `git log` to show for work like this?
- PR body still reads "Work in progress". A paragraph naming the convention and linking EKO-267 would save the next reviewer from having to infer it.

What I actually want to know: did you land on these names by instinct, or was there a specific pattern you were trying to match? If there was a pattern, share it so we can line up. If it emerged as you went, that is worth knowing too — it tells us the convention was implicit already.

Iam-Ebuka:
1.) I agree there’s a small ambiguity with the imported WebSocket type. I leaned toward RealWebSocket because it aligns with the broader convention (Real*, Stubbed*, etc.) and keeps names concise at usage sites. In practice, the distinction becomes clear from context (type vs implementation), but I’m open to keeping RealWebSocketServer if we feel that clarity at the declaration site outweighs consistency.
2.)That’s a fair point about the Client suffix acting as an anchor. I removed it for brevity and consistency with other wrappers, but I agree it introduces some ambiguity around whether “Mongo” refers to the database, driver, or connection. I think this is acceptable given how it’s used in context, but I’m open to keeping Client if we want to prioritize explicitness over consistency.
3.)I think we should document these conventions in a .md file (e.g. CONTRIBUTING.md or a naming conventions doc). Since this PR establishes patterns across the codebase, making them explicit will help future contributors avoid having to infer them from history.

KevinOrfas (approved):

PR #24, opened 2026-04-17
EKO 265 Adopt ConfigurableResponse in FileStorage and ScriptRunner stub

Okpala76:
* **chore: configurable response and emitter implementation for fileStorage**
* **chore: added configurable response to script runner**


KevinOrfas (commented):
Nice direction on this one, it lines up with where Mongo already sits. A few things worth sitting with before we merge.

**On `hasNext()` and the silent fallback**

Open `mongo.ts` and `fileStorage.ts` side by side. In Mongo, when the queue drains, the next call throws. In the new FileStorage, when the queue drains, we quietly fall back to the live store.

Picture this scenario. You write:
```ts
const storage = FileStorage.createNull({ save: [new Error('Disk full')] });
```
Your test expects the first save to fail. Six weeks later I add an extra save in a `beforeEach`. In the lenient version, the error fires on the wrong save, your assertion still passes, we ship a bug and nobody notices. In a strict version, the queue either overflows or runs dry, the test fails, I see the failure, I fix it.

Which world do we want to live in? If we agree on strict, we can drop `hasNext()` entirely and the stub becomes:
```ts
save(name: string, data: Buffer): Promise<void> {
  this.saveResponses?.next();
  this.store.set(name, data);
  this.emitter.emit(CHANGE_EVENT, { type: 'save', name, data });
  return Promise.resolve();
}
```
Same shape as Mongo. One rule across the infrastructure.

**On `StubbedFileSystem.exists`**

Read the two branches. What is different between them? Look again. The emit payload is identical. If they are doing the same thing, can we let them be the same thing?
```ts
const exists = this.existsResponses?.hasNext()
  ? (this.existsResponses.next() as boolean)
  : this.store.has(name);
this.emitter.emit(CHANGE_EVENT, { type: 'exists', name, exists });
return Promise.resolve(exists);
```
Or simpler still once we go strict.

**On `toScriptResult`**

Look at the last line. What stops a test author passing `42` as a response? The type union catches it at compile time today. If someone widens the type later, the cast hides the mistake until a consumer blows up. Worth asking: where would we rather that surface, here at config time, or deep in a caller?

**On naming**

FileStorage has `trackChanges`. ScriptRunner has `trackExecutions`. Mongo has `trackChanges(collection)`. Three verbs, same idea underneath. Is an execution genuinely not a change, or do we want one word across the infrastructure?

**On the signature shift in `ScriptRunner.createNull`**

Old callers wrote `{ echo: 'hello\n' }`. New callers write `{ echo: ['hello\n'] }`. Worth a quick grep across the repo before merge so no caller gets caught out.

**What is working well**

The error path tests carry their weight. The assertion that `tracker.data` stays empty when a configured error fires is exactly the invariant we want pinned down. FileStorage's options shape mirrors Mongo, good consistency. Keeping `trackChanges` as a throw on real instances matches the existing pattern.

**What I would ask before approving**

Just the first one. Strict or lenient on queue exhaustion, and why. Once we agree, the three wrappers should tell the same story.

KevinOrfas:
# EKO-265 review

CI is green: 54 tests pass (9 new), typecheck clean, lint clean aside from the pre existing `console.log` in `start.ts`. Close to merge but a handful of things to clear first, flagged below.

Reading across the four commits: `FileStorage` and `ScriptRunner`'s null backends now share the shape `MongoWrapper` has had for a while — pre seed responses with `ConfigurableResponse`, observe calls with `OutputTracker`. The nine new tests read like behavioural assertions ("throws configured save errors without mutating the store", "does not track executions that throw configured errors") rather than tests that mirror implementation. Tests named that way do a different job: they describe what the wrapper promises its users.

## Before merge

**A small bug in the real backend.** `RealProcessRunner.trackChanges` throws with a message that names a method that does not exist:

```typescript
throw new Error('trackExecutions is only available on null instances');
```

A reader hitting that error would grep for `trackExecutions` and find nothing. One word fix when you're in there.

**Commit hygiene.** All four commits tagged `chore:`. Conventional commits reserves `chore:` for tooling, deps, and build. What this PR adds, configurable responses and output tracking on two nulls, is new functionality, which `feat:` fits better. Also the third commit's message, `removing hasNext lenient buffer to allow for stricter testing`, reads backwards from what the diff does: it adds `hasNext`. Future reader of `git log` will stumble. Worth a quick interactive rebase to reshape the messages before merge.

## Worth sitting with: `ConfigurableResponse` has two open decisions

Both sit on the same class, and the review reads cleaner if they are together.

**Question 1: what happens when the queue is exhausted.** Right now `FileStorage`'s null throws "queue exhausted" if you call an operation more times than you configured responses for. `ScriptRunner`'s null falls back to an empty default (`{ stdout: '', stderr: '', exitCode: 0 }`). Same class, different semantics at the call site. For scripts, a sensible default means tests that don't care about exact output skip the setup. For file storage, strict failure surfaces missing configuration loudly. Both are defensible in their domain. The question is whether this is deliberate. If yes, it would help to name the rule so the next null knows which shape to follow. If it emerged by accident, which shape do we want the other to match? My current lean is strict everywhere, because lenient defaults can quietly mask tests that forgot to configure, but that is a conversation for the pair.

**Question 2: what type the queue holds.** Looking at the class, `unknown` actually starts at `ConfigurableResponse` itself — `queue: unknown[]`, `next(): unknown` — and it propagates up into every wrapper's signature. Which means the fix sits at the base class, not at each consumer. I've got a concrete option to bring to the pair — make `ConfigurableResponse<T>` generic, default `T = unknown` so existing callers keep compiling, and let the wrappers say what they actually accept (`ConfigurableResponse<boolean>` for `exists`, `ConfigurableResponse<void>` for `save`/`remove`, `ConfigurableResponse<ScriptResult | string>` for the runner). The dead `as boolean` cast on `exists` finally has a job, and `save: [true]` stops compiling. Keeping the code for our session rather than pre deciding here. Bring anything you've been thinking about.

The two are independent decisions. Making the class generic does not settle the exhaustion question — a `ConfigurableResponse<boolean>` can still be strict or lenient, that is a consumer choice via `hasNext()` before `next()`. But they belong in one pass rather than two, because both are edits to the same file.

## Merge coordination

This PR still edits `server/infrastructure/output_tracker.ts` (the snake case file). PR #23 is in flight renaming it to `outputTracker.ts`. Whichever merges first creates a conflict for the other. Cheapest path: let EKO-267 merge first since it is smaller and mechanical, then rebase this onto main — the only change is two import paths. Worth pinging the EKO-267 author to agree the order before either of you merges.

## What I'd love your thinking on before we pair

Two questions. First, the strict/lenient call: is the `ScriptRunner` lenient default something you reached for deliberately (tests shouldn't need to configure commands they don't care about), or did it fall out of the port from `Record<string, string>`? Second, once we pair and land on the generic `ConfigurableResponse<T>` shape, should `exists` grow a proper configured return value or stay error only? Bring your reading and I'll bring mine.

KevinOrfas:
Following up on Question 1 from the review above. Options laid out so we have shared material going into the pair.

## The question

When a test seeds a **nullable** (an instance returned by `createNull()`) with fake responses, what should the nullable do if the test runs out of responses but the code under test keeps calling?

A small example to ground it. With `ScriptRunner` today:

```typescript
const runner = ScriptRunner.createNull({
  echo: ['hello\n'],
});

await runner.exec('echo', ['first']);   // returns 'hello\n' (consumed)
await runner.exec('echo', ['second']);  // queue is now empty. What happens?
```

The three wrappers currently answer differently:

- `MongoWrapper`: if you seeded responses, running out throws. If you never seeded any, the nullable returns a sensible default.
- `FileStorage`: same as Mongo.
- `ScriptRunner`: always returns a default, even if you seeded responses and ran out.

We want one rule across all three. Three options.

## Rule A: always throw when out of responses

Every operation your code under test touches must be seeded. Running out, or never seeding at all, both throw.

```typescript
const runner = ScriptRunner.createNull({
  echo: ['hello\n'],
});

await runner.exec('echo', ['first']);   // OK, consumes 'hello\n'
await runner.exec('echo', ['second']);  // throws: queue exhausted
await runner.exec('node', ['x.js']);    // throws: 'node' was never seeded
```

Good: every call is accounted for. Nothing silently passes.
Cost: tests have to configure more than they need to. A test checking `save()` still has to seed responses for any other file system calls the code happens to make, like this:

```typescript
const storage = FileStorage.createNull({
  save: [undefined],     // the call we actually want to assert on
  exists: [false],       // code happens to call exists() first
  remove: [undefined],   // code cleans up an older file
});
```

Even if the test does not care about `exists` or `remove`, it has to seed them to avoid a throw.

## Rule B: always return a default when out of responses

The nullable always has a fallback. Seeding is optional. Running out quietly falls back to the default.

```typescript
const runner = ScriptRunner.createNull({
  echo: ['hello\n'],
});

await runner.exec('echo', ['first']);   // returns 'hello\n'
await runner.exec('echo', ['second']);  // default: { stdout: '', stderr: '', exitCode: 0 }
await runner.exec('node', ['x.js']);    // default
```

Good: tests are short. You only seed what you care about.
Cost: here is the bug this opens up. Start with a test that looks solid:

```typescript
const storage = FileStorage.createNull({
  save: [new Error('disk full')],
});

await expect(storage.save('a.txt', buf)).rejects.toThrow('disk full');
```

Green. Then a colleague adds a retry to the code under test, so `save` is now called twice:

```typescript
// production code under test
try {
  await storage.save(name, buf);
} catch {
  await storage.save(name, buf);   // retry
}
```

Under Rule B, the second `save` runs out of responses and silently returns the default (no error). The original test still passes because `rejects.toThrow('disk full')` only checks the first call. The retry logic is never actually exercised.

A year later someone changes the default. The test breaks. Nobody knows why. That shape of bug is the main thing this rule has to defend against.

## Rule C: throw only when you seeded responses and ran out

If a test seeded responses for an operation, running out throws. If the test never seeded responses for that operation, the nullable runs a default.

```typescript
const runner = ScriptRunner.createNull({
  echo: ['hello\n'],
});

await runner.exec('echo', ['first']);   // OK, consumes 'hello\n'
await runner.exec('echo', ['second']);  // throws: queue exhausted (opted in to echo, ran out)
await runner.exec('node', ['x.js']);    // default (never opted in to node)
```

Good: both sides of the tradeoff get respected. If you chose to care about an operation, you account for every call. If you did not choose to care, you do not pay setup cost to ignore it.
Cost: two states to hold in your head instead of one. A reader has to know whether a given operation is seeded.

## Where I land

Rule C.

Three reasons.

First, `MongoWrapper` and `FileStorage` already follow Rule C. `ScriptRunner` is the one that does not. Moving `ScriptRunner` is a one line change. Moving the other two to either A or B is a much bigger job.

Second, Rule C catches the retry bug from Rule B the moment the second `save` fires, not a year later when someone changes the default.

Third, Rule C does not force setup on tests that do not care. A test about `save` does not have to seed `exists` too.

What would change my mind: if we find real tests in the codebase that deliberately seed one or two responses and then fire more calls against the same operation expecting a default, Rule C would force those tests to be rewritten. If that rewrite feels like the wrong shape, Rule B might be the right call.

## If we agree on Rule C, here is the change

`server/infrastructure/scriptRunner.ts`, inside `StubbedProcessRunner.exec`, one line:

```typescript
// before
exec(command: string, args: string[]): Promise<ScriptResult> {
  const responseQueue = this.responses.get(command);
  const response = responseQueue?.hasNext() ? responseQueue.next() : undefined;
  ...
}

// after
exec(command: string, args: string[]): Promise<ScriptResult> {
  const responseQueue = this.responses.get(command);
  const response = responseQueue !== undefined ? responseQueue.next() : undefined;
  ...
}
```

What this does, in three cases:

```typescript
// 1. Command never configured
const runner = ScriptRunner.createNull();
await runner.exec('anything', []);
// responseQueue is undefined, response stays undefined,
// toScriptResult returns the default.

// 2. Command configured, queue still has items
const runner = ScriptRunner.createNull({ echo: ['hi'] });
await runner.exec('echo', []);
// responseQueue exists, next() returns 'hi',
// toScriptResult turns it into a ScriptResult.

// 3. Command configured, queue exhausted
const runner = ScriptRunner.createNull({ echo: ['hi'] });
await runner.exec('echo', []);   // consumes 'hi'
await runner.exec('echo', []);   // next() throws 'queue exhausted'
```

The existing test `consumes configured responses for repeated command calls before falling back` retires. Two new tests replace it:

```typescript
it('throws when a configured command queue is exhausted', async () => {
  const runner = ScriptRunner.createNull({
    echo: ['hello\n'],
  });

  await runner.exec('echo', ['first']);
  await expect(runner.exec('echo', ['second'])).rejects.toThrow('queue exhausted');
});

it('returns default for commands that were never configured', async () => {
  const runner = ScriptRunner.createNull({
    echo: ['hello\n'],
  });

  await expect(runner.exec('unknown', [])).resolves.toEqual({
    stdout: '',
    stderr: '',
    exitCode: 0,
  });
});
```

## What I want from the pair

If a fourth rule exists that none of A, B, or C captures, bring it. If Rule A or Rule B looks right for reasons I missed, argue for it. If Rule C lands for you, we write the line together.

KevinOrfas:
# PR #24 review: ConfigurableResponse in FileStorage and ScriptRunner

## What works

The "mutation blocked when error configured" invariant is tight. Look at `save` in `StubbedFileSystem` at `server/infrastructure/fileStorage.ts`. The configured response is consumed *before* `store.set` runs. If the response is an error, `next()` throws and the store never mutates. The test `throws configured save errors without mutating the store` pins this down exactly. Same discipline applied in `remove`. Good.

The `trackChanges` event pattern is consistent with how MongoWrapper emits changes, which makes the wrappers feel like siblings rather than cousins. `does not track executions that throw configured errors` confirms that failed calls do not pollute the tracker. That's a subtle thing to get right on the first pass.

Queue exhaustion on `ScriptRunner` (`throws when a configured command queue is exhausted`) is the right default. Fail loud rather than silently return stale data. Copies the pattern from the ConfigurableResponse base class cleanly.

## A real bug worth fixing before merge

`fileStorage.ts`, inside `StubbedFileSystem.exists`:

```ts
exists(name: string): Promise<boolean> {
  this.existsResponses?.next() as boolean;
  const exists = this.store.has(name);
  ...
  return Promise.resolve(exists);
}
```

Two things are going on here that shouldn't be.

First, `this.existsResponses?.next() as boolean;` is an expression with no effect. The return value of `next()` is read and immediately discarded. The `as boolean` cast does nothing because nothing is being assigned. Either remove the cast or use the value.

Second, whatever `next()` returns is thrown away. So if a caller writes `createNull({ exists: [true] })` hoping to override what `store.has` says, their `true` goes in the bin. The store's real state wins. Only the *error throwing* side of ConfigurableResponse survives.

Two ways to resolve it. Pick one and test it.

**Option A: errors only, no value override.** Intentional. Then the type should say so: `exists?: Error[]` instead of `exists?: unknown[]`, and drop the dead `as boolean` cast. Document in the test that exists values cannot be overridden.

**Option B: allow value override.** Use the response: `const override = this.existsResponses?.next(); const exists = typeof override === 'boolean' ? override : this.store.has(name);`. Add a test for `createNull({ exists: [true] })` against an empty store.

Worth asking: which did you mean?

## Question for the author

The `save` and `remove` methods emit `CHANGE_EVENT` with `type: 'save'` or `type: 'remove'`. The `exists` method also emits with `type: 'exists'`. Fine on its own, but `exists` is a *read*, not a *change*. Is a read hitting the `change` channel the right model, or does reading belong on a separate tracker (`trackReads`)? Not a blocker, but worth a minute of thought before the pattern gets copied into the next wrapper.

## Merge conflict on the horizon

This branch is based on main before EKO-267 landed. Current main has `FileStorageWrapper`, `FileStorageInterface`, `ScriptRunnerWrapper`, `ScriptRunnerInterface`, `outputTracker.ts`. This PR still uses `FileStorage`, `FileSystemInterface`, `ScriptRunner`, `ProcessRunnerInterface`, `output_tracker.ts`. Rebase onto main will touch every file in the diff. Do the rebase yourself rather than let the merge UI guess.

## Small things, non blocking

The `ScriptRunnerResponses` type accepts a single response or an array of responses per command (`ScriptRunnerResponse | ScriptRunnerResponse[]`). The tests only ever pass arrays. The single-value shorthand is untested. Either add a test for it or drop the shorthand to keep the surface small.

The `toScriptResult` helper at the bottom of `scriptRunner.ts` handles three cases: string, undefined, ScriptResult. The string and undefined branches converge on the same shape. You could collapse them with a default and a conditional. Stylistic, yours.

## What I would do

Fix the `exists` bug. Rebase onto main. Answer the read-vs-change question in the PR thread. Then this is ready.

PR #25, opened 2026-04-20
EKO-260: Refactor createServer to accept WebSocketWrapper                                                                                              

KevinOrfas:
## Summary
- Removes `server/plugins/websocketRoutes.ts` in favour of injecting `WebSocketWrapper` into `createServer`
- Updates `server/start.ts` to construct the wrapper and pass it in
- Adds a nullable test that exercises the server wiring without a real socket

Linear: https://linear.app/ekohacks/issue/EKO-260

## Test plan
- [x] `npm test`
- [x] `npm run test:integration`
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] Boot with `npm run dev:server` and open a WebSocket to `ws://localhost:3001/ws` in the browser console


KevinOrfas:
# EKO-260 review

The state of the branch, the contradiction in the story, and what it asks of all of us.

## What's broken

`server/start.ts` builds a Fastify instance, hands it to the wrapper, then calls `createServer` which builds a **second** Fastify internally. The wrapper registers `/ws` on the first. `server.listen(3001)` listens on the second. Run `npm run dev:server` and open `ws://localhost:3001/ws` in a browser console. Nothing connects.

The wrapper integration test is green because it wires its own Fastify end to end and never touches `createServer`. The one test that did go through `createServer` now uses `createRawWs({ port: 1010 })`, which fails with `EACCES` because 1010 is privileged on Linux. Even if it ran, it starts a standalone WS server on 1010 while the client connects to Fastify's dynamic port. Two different servers. The test would not prove what it claims.

The nullable test was added, but it lives in `tests/infrastructure/websocket.test.ts`, a wrapper unit test file, while it exercises `createServer`. Wrong address.

Commit: `chore: test websocket imleemattion`. Typo, and not a chore.

## Where this came from

Read the story's Step 1 and Step 2 side by side:

```typescript
// Step 1
export async function createServer(options: ServerOptions) {
  const server = Fastify();              // server creates its own
  ...
}
```

```typescript
// Step 2
const fastify = Fastify();
const ws = WebSocketWrapper.create(fastify);
await createServer({ ws });
await fastify.listen({ port: 3001 });    // listens on the one start.ts made
```

The two snippets cannot both be right. If `createServer` makes its own Fastify, the one in `start.ts` is orphan. If `start.ts`'s Fastify is the real one, `createServer` must receive it, not create it.

You typed the sample as written. The sample was wrong. I wrote the story and did not catch my own contradiction. Whoever reviewed the story before it became EKO-260 did not catch it either. We are all in this.

## What this tells us about the process

A planning doc that contradicts itself made it into the branch. That is not one person's mistake. It is a gap in how we write and review EkoLite stories. Real questions, not rhetorical:

- When we write a story like this, who reads it before it becomes a Linear ticket?
- If you had spotted the contradiction while committing, would you have felt safe pausing to flag it, or would the story's authority have carried you through?
- Should EkoLite stories be reviewed the way code is, with someone other than the writer required to sign off?

Bring these to the next retro. They matter more than the fix.

## The real design question

"Who owns the Fastify instance" is the decision underneath the bug. Two honest answers. Neither is obvious.

**Option A: the caller owns Fastify.** `createServer` accepts it, registers static on it, `ws.start()` registers `/ws` on the same instance, `start.ts` calls `fastify.listen()` itself.
- Good: one instance, wrapper and server share it, no orphans possible.
- Bad: every caller constructs Fastify. More boilerplate. `FastifyInstance` leaks into any test that just wants a server.

**Option B: `createServer` owns Fastify and returns it.** The wrapper is created inside, or `createServer` accepts a factory for it.
- Good: callers never touch Fastify.
- Bad: the wrapper's `create(fastify)` factory needs a Fastify that doesn't exist yet. You either change the wrapper, or let `createServer` build the wrapper too, which kills the injection the story set out to introduce.

The story pushed toward a third option that doesn't hold up: caller owns the wrapper, `createServer` owns Fastify. That is the contradiction, made concrete.

Pick A or B knowing what each costs. There is no "right" one without the cost.

## What happens next

Not homework from me to you. Shared work:

- We pair on rewriting `createServer` with A or B, whichever you want to argue for.
- I fix the story so the next person reading it is not set up to fail the way you were.
- You bring the process questions to the retro so the fix is not just code.

KevinOrfas:
# EKO-260 review, part two: the design question

The first comment names what broke and where it came from. Before anyone writes code, we sit with the question the broken sample was trying to answer: **who owns the Fastify instance?**

Two options. Both work. Both cost something. The shape we pick depends on what we value.

## Option A — caller owns Fastify

Shape: `createServer(fastify, { ws })`. The caller constructs Fastify and the wrapper, then passes both in. `createServer` configures the Fastify it received.

```typescript
// server/index.ts
export async function createServer(fastify: FastifyInstance, options: { ws: WebSocketWrapper }) {
  await fastify.register(fastifyStatic, { root: resolve(__dirname, '..', 'dist', 'client') });
  await options.ws.start();
  return fastify;
}

// server/start.ts
const fastify = Fastify();
const ws = WebSocketWrapper.create(fastify);
await createServer(fastify, { ws });
await fastify.listen({ port: 3001 });
```

Tradeoffs:
- Fastify is named and visible in every entry point and every test that boots the server.
- `createServer` becomes configurational rather than constructive. If the team minds, it renames to `configureServer`.
- Adding more infrastructure: the options bag grows to `{ ws, mongo, fileStorage }`. Callers construct each dependency directly.
- The framework choice lives in every caller.

## Option B — createServer owns Fastify, accepts a wrapper factory

Shape: `createServer({ wsFactory })` returns `{ fastify, ws }`. The caller passes a factory function, `createServer` uses it to build the wrapper internally.

```typescript
// server/index.ts
export async function createServer(options: { wsFactory: (f: FastifyInstance) => WebSocketWrapper }) {
  const fastify = Fastify();
  const ws = options.wsFactory(fastify);
  await fastify.register(fastifyStatic, { ... });
  await ws.start();
  return { fastify, ws };
}

// server/start.ts
const { fastify } = await createServer({ wsFactory: (f) => WebSocketWrapper.create(f) });
await fastify.listen({ port: 3001 });
```

Tradeoffs:
- Fastify is hidden inside `createServer`. Callers never import it.
- The caller passes a function instead of an instance.
- Adding more infrastructure: the options bag grows to `{ wsFactory, mongoFactory, fileStorageFactory }`. Callers write a factory for each dependency.
- The framework choice lives in one place. If we ever swap Fastify out, the change is contained.

## The questions we cannot answer alone

Both options pass tests. Both fix the bug. The choice lives upstream of the code.

- What is EkoLite optimising for right now — the readability of the injection pattern across our own codebase, or the portability of that pattern to other frameworks in future?
- How likely is it we swap Fastify for another framework in the next year or two? If it is unlikely, the encapsulation Option B offers has less weight. If it is on the roadmap, it has more. What do we actually believe about that roadmap?
- The wrapper currently takes a Fastify instance. Option A continues that shape through to `createServer`. Option B introduces a factory layer that keeps the wrapper the same but hides Fastify from the caller. Which shape does the team find easier to reason about?
- When Mongo and FileStorage arrive, Option A asks callers to construct each dependency. Option B asks callers to write a factory for each. Which do we prefer to write and read across many such wrappers?
- Is there an Option C here that neither A nor B captures?

## What happens next

Read this. If you lean toward A, make the case for A. If toward B, make the case for B. If you see a third option, name it. Put your reasoning on this PR or on the Linear ticket. Once both arguments are written, we rewrite the code and the story together.

KevinOrfas:
# EKO-260: two Fastify instances, and the test that caught it

## What broke

`server/index.ts` called `Fastify()` inside `createServer()`. `server/start.ts` called `Fastify()` again outside, then passed that one to `WebSocketWrapper.create(fastify)`. The wrapper registered `/ws` on the outer Fastify. `server.listen(3001)` ran on the inner one. The route and the listener were on different objects.

Reproduce it on the branch before this fix:

```
npm run dev:server
# in browser devtools:
new WebSocket('ws://localhost:3001/ws')
# close 1006, no route
```

Same command on this branch connects.

## What the test was asking

`tests/server/server.integration.test.ts`:

```ts
const ws = WebSocketWrapper.create();
server = await createServer({ ws });
await server.listen({ port: 0 });
const port = String(server.addresses()[0].port);
client = new WebSocket(`ws://localhost:${port}/ws`);
```

Read it carefully. `create()` takes no argument. At that point in the flow, the caller has no Fastify to pass, because `createServer` hasn't built one yet.

The old signature was `create(fastify: FastifyInstance)`. It demanded a value nobody had. That gap is why `start.ts` invented a throwaway Fastify, and why the `/ws` route ended up orphaned.

Question for you. If a test compiles against an API that doesn't exist yet, which side do you change? We changed the API. Why?

## The shape we landed on

Two methods on `WebSocketWrapper`, each tied to a mode:

```ts
static create(): WebSocketWrapper
async attach(fastify: FastifyInstance): Promise<void>   // Fastify mode
async start(): Promise<void>                             // raw mode
```

`createServer` now calls `await options.ws.attach(server)` after it builds its Fastify. The Fastify that listens is the same object the route gets registered on. Single instance, single lifecycle.

Raw mode (`createRawWs({port})`) uses `start()` to boot a standalone `WebSocketServer` on its own port. Fastify mode (`create()`) uses `attach(fastify)` to bind to an existing Fastify. Two modes, two verbs. They are not interchangeable.

## Where the bug came from

The spec had Step 1 saying `createServer` creates its own Fastify, and Step 2 saying `start.ts` creates the Fastify that gets listened on. Both can't be true. Nobody flagged it at spec review, at ticket creation, or while the branch was being written.

Retro questions:

- Who reads a spec before it becomes a Linear ticket?
- If a sample in a spec contradicts itself, at what point in the flow is it cheapest to flag?
- Should specs get a second reviewer the way code does?

## Pattern to carry forward

When you inject a dependency, check the timing. If the dependency doesn't exist at the moment you'd construct the thing that needs it, splitting construction from binding is usually the cheaper fix. `create()` plus `attach(x)` beats `create(x)` when `x` isn't available yet.

Iam-Ebuka (approved):
This is good

PR #26, opened 2026-04-22
Eko-232 server side publication

Iam-Ebuka:
- Initialize publication.
- wrote test "sends error when subscribing to unknown publication".
- wrote implementation for the test.

https://linear.app/ekohacks/issue/EKO-232/3a1-register-a-publication

### DIFF
--- a/server/logic/publications.ts
+++ b/server/logic/publications.ts
@@ -9,14 +9,34 @@ interface PublicationInterface {
 
 type PublicationDef = () => { collection: string; query: object };
 
-export class Publications implements PublicationInterface {
+export class PublicationsWrapper {
+  private client: PublicationInterface;
+
+  private constructor(client: PublicationInterface) {
+    this.client = client;
+  }
+
+  static createNull(mongo: MongoWrapper, ws: WebSocketWrapper): PublicationsWrapper {
+    return new PublicationsWrapper(new StubbedPublication(mongo, ws));
+  }
+
+  define(name: string, queryFn: PublicationDef): void {
+    this.client.define(name, queryFn);
+  }
+
+  async handleMessage(clientId: string, message: ClientMessage): Promise<void> {
+    return this.client.handleMessage(clientId, message);
+  }
+}
+
+class StubbedPublication implements PublicationInterface {
   private publications = new Map<string, PublicationDef>();
   private ws: WebSocketWrapper;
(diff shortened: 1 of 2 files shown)


KevinOrfas:
Nice one Ebuka, the story brief lines up with what you've shipped. The async signature is in place for 3.A.2, mongo and ws both on the constructor so nothing needs reshuffling next round, error shape matches `ErrorMsg`, and you've held off on unsubscribe and the happy path like the story asks.

One thing I'd like us to chew on together before you start 3.A.2, and that's `PublicationInterface`.

The story didn't ask for it, and the pattern we're using for interfaces in this codebase is a bit different from where you've put this one. Have a look at `MongoInterface` and `WebSocketInterface`. Each has a real implementation and a stubbed implementation, and the wrapper picks between them at construction. The interface earns its keep there because it's the contract two sides have to honour, and swapping them is how the nulled infrastructure pattern works.

`Publications` sits a layer above that. No real or null variants. It's pure domain logic that takes nulled infrastructure in through its constructor. `PublicationInterface` in this file has exactly one implementer, `Publications` itself, and the two shapes are going to have to move together every time the class grows.

A few questions worth sitting with:

- What do you hope `PublicationInterface` is doing for you? If it's close to "documenting the public API of the class", what's the class signature already doing?
- Trace how `MongoInterface` is actually used from construction through to call site. Where does it stop being decorative and start being load bearing?
- If you deleted `PublicationInterface` from this file right now, which test would fail?

No right answer I'm nudging you toward. If after working through those you still want to keep it, I'd just want to hear the reasoning, because the class will pick up three more behaviours across 3.A.2 to 3.A.4 and the interface will have to grow with it every time.

Tiny thing while we're here: inside an `async` function a plain `return` already gives you a resolved `Promise<void>`, so the two `return Promise.resolve()` lines in `handleMessage` can just be `return`. Cosmetic, not important.

On 3.A.2, the happy path will drive `mongo`. Worth reading `ConfigurableResponse` in `server/infrastructure/outputTracker.ts` before you pick it up, because that's how `MongoWrapper.createNull({ find: [...] })` feeds seeded results back, and it's easier to work with once you've seen the shape of it.

KevinOrfas (commented):
@Iam-Ebuka Approved, just a couple of questions to sit with for the following PRs

PR #28, opened 2026-04-22
refactor: clean up to maintain Rule C

Okpala76:
This branch refactors 

The use has next and aligns with Rule C( link below for context)

https://github.com/ekohacks/ekolite/pull/24#issuecomment-4282849294

### DIFF
--- a/server/infrastructure/outputTracker.ts
+++ b/server/infrastructure/outputTracker.ts
@@ -54,10 +54,6 @@ export class ConfigurableResponse {
     this.queue = [...responses];
   }
 
-  hasNext(): boolean {
-    return this.queue.length > 0;
-  }
-
   next(): unknown {
     if (this.queue.length === 0) {
       throw new Error('ConfigurableResponse queue exhausted — no more responses configured');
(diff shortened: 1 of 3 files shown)


KevinOrfas:
Nice one Gochi. The behavioural change is exactly what the ticket asked for.

The one line in `StubbedProcessRunner.exec` matches the spec. The old `consumes configured responses for repeated command calls before falling back` test is retired because that was the Rule B assertion. The two new tests land with the bodies straight out of the PR #24 follow up comment, and CI is green. `MongoWrapper`, `FileStorageWrapper`, and `ScriptRunnerWrapper` are now all Rule C.

One thing you did that went beyond the ticket, which I like: you deleted `hasNext()` from `ConfigurableResponse`. Nothing else in the codebase used it, and leaving it there would have left the door open for the next person to reach for the same lenient pattern you just removed. Worth hearing your reasoning though, because what you did is a design decision worth being explicit about. Was it "this is dead code, delete it", or "this method enables the anti pattern, delete it so nobody can reach for it that way again"? Both are fine answers, they just suggest slightly different things about when you'd add a method back.

One thing from the Definition of Done that didn't make it into this PR: the PR description currently reads "This branch refactors the use has next and aligns the to rule C". When someone reads this PR in six months, they'll want a link to the PR #24 follow up comment where Rule C was reasoned through. One line and an anchor link is enough. The reasoning trail is half the value of the ticket.

Happy to approve once the PR description is updated.

KevinOrfas (commented):
Approved, please link the PR #24 for visibility and maintenance

PR #29, opened 2026-04-23
Eko 231/3a2 send initial documents on subscribe

Iam-Ebuka:
- test for send initial documents on subscribe
- enhanced handleMessage in publication for handling existing publications
- test for sends ready even when no documents exist
- red -> green -> refactor was followed

https://linear.app/ekohacks/issue/EKO-231/3a2-send-initial-documents-on-subscribe

### DIFF
--- a/server/logic/publications.ts
+++ b/server/logic/publications.ts
@@ -33,7 +33,6 @@ export class Publications {
   async handleMessage(clientId: string, message: ClientMessage): Promise<void> {
     if (message.type === 'subscribe') {
       const queryFn = this.publications.get(message.name);
-      const publicationsExists = this.publications.has(message.name);
 
       if (!queryFn) {
         this.ws.send(clientId, {
@@ -44,15 +43,13 @@ export class Publications {
         return Promise.resolve();
       }
 
-      if (publicationsExists) {
-        const { collection, query } = queryFn();
-        const docs = await this.mongo.find<{ _id: string }>(collection, query);
+      const { collection, query } = queryFn();
+      const docs = await this.mongo.find<{ _id: string }>(collection, query);
 
-        for (const doc of docs) {
-          this.ws.send(clientId, toAddedMsg(collection, doc));
-        }
-        this.ws.send(clientId, toReadyMsg(message.id));
+      for (const doc of docs) {
+        this.ws.send(clientId, toAddedMsg(collection, doc));
       }
+      this.ws.send(clientId, toReadyMsg(message.id));
     }
     return Promise.resolve();
   }
--- a/server/logic/publications.ts
+++ b/server/logic/publications.ts
@@ -4,6 +4,18 @@ import { WebSocketWrapper } from '../infrastructure/websocket.ts';
 
 type PublicationDef = () => { collection: string; query: object };
 
+const toAddedMsg = (collection: string, doc: Record<string, unknown>) => ({
+  type: 'added',
+  collection,
+  id: doc._id as string,
+  fields: Object.fromEntries(Object.entries(doc).filter(([key]) => key !== '_id')),
+});
+
+const toReadyMsg = (subId: string) => ({
+  type: 'ready',
+  id: subId,
+});
+
 export class Publications {
   private publications = new Map<string, PublicationDef>();
   private ws: WebSocketWrapper;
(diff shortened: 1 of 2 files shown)


server/logic/publications.ts
```diff
@@ -29,6 +43,16 @@ export class Publications {
         });
         return Promise.resolve();
       }
+
+      if (publicationsExists) {
```
KevinOrfas: Once the `!queryFn` early return fires, `publicationsExists` can only be true here, so both the `has()` call on line 36 and this `if` guard are dead. I would drop `publicationsExists` entirely and let the body run unconditionally. If there is a case you had in mind where `get` returns truthy but you still want a second gate, let me know, otherwise safe to remove.

tests/logic/publications.test.ts
```diff
@@ -22,4 +22,52 @@ describe('Publications', () => {
...
+    expect(client.messages).toContainEqual({
+      type: 'added',
+      collection: 'files',
+      id: '1',
+      fields: { name: 'existing.bam' },
+    });
+    expect(client.messages).toContainEqual({
+      type: 'ready',
+      id: 'sub1',
+    });
```
KevinOrfas: Two `toContainEqual` calls will pass even if `ready` arrives before `added`, but DDP semantics require initial `added`s to flush before `ready`. The empty-collection test just below pins the contract tightly with `toHaveLength` + `toEqual` — could this one do the same shape?\n\n```ts\nexpect(client.messages).toEqual([\n  { type: 'added', collection: 'files', id: '1', fields: { name: 'existing.bam' } },\n  { type: 'ready', id: 'sub1' },\n]);\n```\n\nThat way an ordering regression or an accidental duplicate `added` shows up in the test.

KevinOrfas (commented):
A couple of tweaks before this is ready to land. Unsubscribe race is out of scope (3a3 picks it up).

server/logic/publications.ts
```diff
@@ -4,6 +4,18 @@ import { WebSocketWrapper } from '../infrastructure/websocket.ts';
 
 type PublicationDef = () => { collection: string; query: object };
 
+const toAddedMsg = (collection: string, doc: Record<string, unknown>) => ({
+  type: 'added',
+  collection,
+  id: doc._id as string,
+  fields: Object.fromEntries(Object.entries(doc).filter(([key]) => key !== '_id')),
+});
```
KevinOrfas, line 12: Not a change for this PR, just a seed for next time. `added` is one of three data message variants in `DataMsg` (alongside `changed` and `removed`). When those land, you will likely end up with `toChangedMsg` and `toRemovedMsg`, three helpers differing by one string field. At that point a single `dataMessage(type, collection, doc)` might read better than three near-twins. Worth holding in mind as the protocol grows, so the duplication does not surprise you. 

Also the `to` prefix reads like a conversion (think `toString`, `toJSON`). Here you are constructing, not converting. Something like `addedMessage` or `buildAddedMessage` might be closer to what it actually does. Small thing, feel free to disagree.

KevinOrfas (commented):
One more to sit with, not a blocker.

server/logic/publications.ts
Iam-Ebuka: Good catch looking ahead. Although I do not understand the second paragraph clearly

KevinOrfas (approved):
Both pieces landed cleanly. `handleMessage` now reads linearly, and the happy path test pins count and order, mirroring the empty-collection case. Leaving the naming question as a seed for when `changed` and `removed` arrive was the right call. Nice work.

server/logic/publications.ts
KevinOrfas: Fair, let me unpack it.

The JavaScript convention is that names starting with `to` describe a conversion: something comes in, a different representation comes out. `date.toISOString()` turns a Date into a string. `JSON.stringify` (or a `toJSON` method) turns an object into a JSON string. `array.toSorted()` turns an array into a new sorted array. The input and output are two shapes of the same thing.

What `toAddedMsg` does is different. Nothing is being converted. You are passing in a collection name and a document, and getting back a brand new object that did not exist before, a protocol message. That is construction, not conversion.

So the reader sees `toAddedMsg` and their brain reaches for "what is being turned into a message". The answer is "nothing really", which is a small moment of friction. A name like `addedMessage(collection, doc)` or `buildAddedMessage(collection, doc)` describes the act more honestly, you are building a message.

It is a micro point about reading flow, not correctness. The code works either way. Just wanted to share the instinct so you can weigh it next time you reach for a `to` prefix.

Iam-Ebuka: Thanks for the clarification, this is helpful

PR #30, opened 2026-04-23
Eko 230/3a3 push live changes

Iam-Ebuka:
This is golden

- red test for push live changes
- wrote the implementation to watch for live changes
- refactor, introduced isChangeEvent function into watchChanges in mongo to maintain intergrity

### DIFF
--- a/server/infrastructure/mongo.ts
+++ b/server/infrastructure/mongo.ts
@@ -1,5 +1,5 @@
 import { Db, MongoClient as Driver, ObjectId } from 'mongodb';
-import { ChangeEvent } from '../../shared/types.ts';
+import { ChangeEvent, isChangeEvent } from '../../shared/types.ts';
 import { ConfigurableResponse, EventEmitter, OutputTracker } from './outputTracker.ts';
 
 interface MongoInterface {
@@ -7,9 +7,8 @@ interface MongoInterface {
   insert(collection: string, doc: object): Promise<void>;
   update(collection: string, query: object, changes: object): Promise<void>;
   remove(collection: string, query: object): Promise<void>;
-  watchChanges(collection: string, cb: (data: unknown) => void): void;
+  watchChanges(collection: string, cb: (change: ChangeEvent) => void): void;
   trackChanges(collection: string): OutputTracker;
-  
 }
 
 export class MongoWrapper {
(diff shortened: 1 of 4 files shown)


server/infrastructure/mongo.ts
```diff
@@ -160,6 +169,12 @@ class StubbedMongo implements MongoInterface {
     return Promise.resolve();
   }
 
+  watchChanges(collection: string, cb: (change: ChangeEvent) => void): void {
+    this.emitter.on(collection, (data) => {
+      if (isChangeEvent(data)) cb(data);
+    });
+  }
```
KevinOrfas, line 176: **What we saw**

`StubbedMongo.watchChanges` was telling two stories at once. The signature promised callers `(change: ChangeEvent) => void`, but the body handed the callback straight to an emitter that calls handlers with `unknown`. The compiler caught the mismatch, and we silenced it with `cb as (data: unknown) => void`. That cast looks small. It isn't. It's us telling the type system "trust me" while the evidence says otherwise.

**What the evidence said**

The same emitter carries more than change events. `find` calls emit `{ type: 'find', collection, query }` on the same channel, because `trackChanges` (our test-time output tracker) needs to see them. So a `watchChanges` subscriber who trusts the declared type would eventually meet a `find` payload that doesn't fit the `ChangeEvent` union. Today we happen to dodge it — `publications.ts` only reacts to `change.type === 'insert'` and silently drops everything else. Dodging isn't the same as handling. The next caller who writes an exhaustive `switch` on `change.type` inherits a landmine.

**The move**

Instead of lying to the type system, we made the boundary honest. The handler we register with the emitter now has the type the emitter actually calls it with — `(data: unknown) => void`. Inside, a small guard (`isChangeEvent`, living next to the type it describes in `shared/types.ts`) narrows `unknown` down to `ChangeEvent` before the caller's `cb` ever sees it. Non-change payloads like `find` are filtered out where they enter, not where they fail.

```ts
watchChanges(collection, cb) {
  this.emitter.on(collection, (data) => {
    if (isChangeEvent(data)) cb(data);
  });
}
```

Two things changed, one thing didn't. The cast is gone. The callback contract is now enforced at runtime, not asserted at compile time. The public shape of `watchChanges` is identical — no caller had to change.

**Why it holds**

The guard is the single place that decides what counts as a `ChangeEvent`. If we ever add a fourth variant to the union, the guard is the one place that has to keep up — and the compiler will tell us, because `isChangeEvent` uses the union's discriminants directly. That's the invariant we're buying: the claim on the signature matches what the callback actually receives, and there's exactly one place to check when the claim shifts.

**What we left open**

`emitter.on` still returns `void`, which means a registered watcher can't be removed. We're not using teardown yet — no unsubscribe, no disconnect cleanup — so it doesn't hurt us today. It will, the first time a client reconnects and we notice the old listener still firing into a dead socket. That's a conversation for the slice that introduces `unsubscribe`, not this one. Worth knowing it's waiting.

PR #31, opened 2026-04-24
Eko 229/3a4 unsubscribe stops updates

Iam-Ebuka:
This is as done as a fine Lambo

- test - red unsubscribed clients don't update
- test - green, implemented the code for it
- refactored, mongo.ts watchChanges method, emitter in outputTracker.ts, publications.ts to track and delete unsubscribed user

https://linear.app/ekohacks/issue/EKO-229/3a4-unsubscribe-stops-updates

### DIFF
--- a/server/infrastructure/mongo.ts
+++ b/server/infrastructure/mongo.ts
@@ -7,7 +7,7 @@ interface MongoInterface {
   insert(collection: string, doc: object): Promise<void>;
   update(collection: string, query: object, changes: object): Promise<void>;
   remove(collection: string, query: object): Promise<void>;
-  watchChanges(collection: string, cb: (change: ChangeEvent) => void): void;
+  watchChanges(collection: string, cb: (change: ChangeEvent) => void): () => void;
   trackChanges(collection: string): OutputTracker;
 }
 
@@ -49,8 +49,8 @@ export class MongoWrapper {
     return this.client.remove(collection, query);
   }
 
-  watchChanges(collection: string, cb: (change: ChangeEvent) => void): void {
-    this.client.watchChanges(collection, cb);
+  watchChanges(collection: string, cb: (change: ChangeEvent) => void): () => void {
+    return this.client.watchChanges(collection, cb);
   }
 
   trackChanges(collection: string): OutputTracker {
@@ -82,7 +82,7 @@ class RealMongo implements MongoInterface {
     await this.db.collection(collection).deleteMany(query);
   }
 
-  watchChanges(_collection: string, _cb: (change: ChangeEvent) => void): void {
+  watchChanges(_collection: string, _cb: (change: ChangeEvent) => void): () => void {
     throw new Error('watchChanges is not implemented for RealMongo');
   }
 
@@ -169,10 +169,14 @@ class StubbedMongo implements MongoInterface {
     return Promise.resolve();
   }
 
-  watchChanges(collection: string, cb: (change: ChangeEvent) => void): void {
-    this.emitter.on(collection, (data) => {
+  watchChanges(collection: string, cb: (change: ChangeEvent) => void): () => void {
+    const listener = (data: unknown) => {
       if (isChangeEvent(data)) cb(data);
-    });
+    };
+    this.emitter.on(collection, listener);
+    return () => {
+      this.emitter.off(collection, listener);
+    };
   }
 
   trackChanges(collection: string): OutputTracker {
--- a/server/infrastructure/mongo.ts
+++ b/server/infrastructure/mongo.ts
@@ -1,5 +1,5 @@
 import { Db, MongoClient as Driver, ObjectId } from 'mongodb';
-import { ChangeEvent } from '../../shared/types.ts';
+import { ChangeEvent, isChangeEvent } from '../../shared/types.ts';
 import { ConfigurableResponse, EventEmitter, OutputTracker } from './outputTracker.ts';
 
 interface MongoInterface {
@@ -7,9 +7,8 @@ interface MongoInterface {
   insert(collection: string, doc: object): Promise<void>;
   update(collection: string, query: object, changes: object): Promise<void>;
   remove(collection: string, query: object): Promise<void>;
-  watchChanges(collection: string, cb: (data: unknown) => void): void;
+  watchChanges(collection: string, cb: (change: ChangeEvent) => void): void;
   trackChanges(collection: string): OutputTracker;
-  
 }
 
 export class MongoWrapper {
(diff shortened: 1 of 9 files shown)


KevinOrfas:
## Review — Story 3.A.4: unsubscribe stops updates

The red/green/refactor cadence is visible in the commit trail and the tests read cleanly against the null Mongo. The story is scoped to "unsubscribe tears down the watcher", and the PR delivers that plus the two tests the story asks for. Good shape to open a conversation on.

Below are the things I'd want us to think through together before merging. Some are in scope for 3.A.4, some are signposts for where 3.A might want to land.

### Worth resolving in this PR

**The story's own watch-out: track subscriptions per client.**
The story flags this directly: *"A client might have multiple active subscriptions. Use a key like `${clientId}:${subId}` or a nested Map. Don't assume one subscription per client."* The current `this.subscriptions.set(message.id, cleanup)` in `publications.ts:55` keys on `message.id` alone. Sub IDs are client-chosen, so two clients can both pick `'sub1'`, and whichever subscribes second silently stomps the first's cleanup reference. What do you reckon about composite keys, or a `Map<clientId, Map<subId, cleanup>>`? A two-client test would lock it in.

**What if the same sub ID gets reused on one client?**
Same site: if a client sends `subscribe` twice with `id: 'sub1'` without unsubscribing in between, the second overwrite drops the first cleanup on the floor. Worth deciding whether that's a protocol error, an idempotent no op, or a "dispose the old one first" path. Curious which feels right to you.

**The `isChangeEvent` guard is looser than the new union promises.**
`shared/types.ts` now says insert and update carry `fields: Record<string, unknown>` (no longer optional). The guard only checks `type`. A malformed event with `type: 'insert'` and no `fields` would pass, and `addedMessage(collection, change.fields)` would then read `undefined` at runtime behind a type that says otherwise. Two ways out: validate `fields` / `id` / `collection` in the guard, or keep `fields` optional on the union and handle absence in the consumer. Either's fine, whichever matches how you want the boundary to feel.

### Worth a sanity check

**`EventEmitter.emit` iterates the live array while `off` splices it.**
In `outputTracker.ts`, `emit` walks `handlers` with `for...of` at the same time `off` can splice it. Nothing today removes itself mid emit, but a snapshot (`for (const h of [...handlers])`) is cheap insurance and matches what people expect from a pub/sub. Up to you whether that's worth doing here or later.

**How strongly does the "no leak" test actually prove no leak?**
It asserts no extra messages arrive, which is also true if the map hangs on to stale entries whose cleanups ran. The story endorses exactly this shape of test so this isn't a block, but if you wanted a sharper version you could expose `subscriptions.size` through a test seam and assert it's zero after the loop. Open question whether that's worth it or overkill for the story.

### Out of scope, worth noting

**No teardown on client disconnect.**
If a client drops the socket without sending `unsubscribe`, its watcher lives forever. The story explicitly scopes to the unsubscribe message, so this isn't a 3.A.4 concern, but it's the same teardown mechanism and probably wants a Linear ticket so it doesn't get forgotten before the next story lands.

### Small things
- `RealMongo.watchChanges` throws with `'watchChanges is not implemented for RealMongo'` while the neighbouring `trackChanges` throws with `'trackChanges is only available on null instances'`. Not wrong, just two different registers sitting next to each other.
- Commit trail is clean and the TDD rhythm is legible from the outside. Nice.

### Over to you

The composite key question is the one I'd most like to hear your thinking on, since the story calls it out and it's the one thing that changes behaviour the moment a second client joins. Happy to pair on that or re-read once you've decided which direction to take.

KevinOrfas:
## Second pass review

Per client `Map<clientId, Map<subId, cleanup>>` is the right shape, dispose old first is the right default for repeat subscribes, and the `RealMongo.watchChanges` error wording now matches `trackChanges`. Good.

One thing to sort before this merges. I'll be specific about what to look at.

### Step 1: a test that looks like it works

Take your "stops sending updates after unsubscribe" test and copy it. In the copy, instead of sending the `unsubscribe` message, call `client.close()`. `StubbedClient` has a `close()` method that disconnects through the wrapper.

Run it. It passes. The assertion `expect(newMessages).toHaveLength(0)` is satisfied because no extra messages arrived after `client.close()`.

The question worth asking is whether "passes" here means "the watcher was torn down", or whether the test is satisfied for some other reason.

### Step 2: read what `send` does to a disconnected client

Open `server/infrastructure/websocket.ts` and find `StubbedWebSocket.send(clientId, message)`. It does a Map lookup on `clients`. If the client is not in the map, the `if (client)` check is false, and `send` returns without doing anything.

Now trace the path through the step 1 test:

1. `subscribe` sets up a watcher on the Mongo emitter.
2. `client.close()` removes the client from `StubbedWebSocket`'s map.
3. You insert a document. The Mongo change event fires.
4. The watcher callback runs, because it is still registered, and calls `ws.send(clientId, addedMessage(...))`.
5. The client is not in the map, so `send` silently no-ops.
6. Your test sees zero new messages. Pass.

Notice what just happened. The test passes whether or not the watcher was cleaned up. The watcher is still registered on the Mongo emitter, still doing work, still calling a `send` that goes nowhere. The test cannot see the difference between "watcher was torn down" and "watcher is alive but its outputs go to a client that no longer exists".

So step 1 is the wrong test for the thing we actually care about.

### Step 3: does the server even know the client left?

Look at three places in `server/infrastructure/websocket.ts`:

- `StubbedWebSocket.disconnect(clientId)`. Deletes from the map, then emits `DISCONNECTION_EVENT` on the internal emitter.
- `RealWebSocket.start()`, inside the `socket.on('close', ...)` handler. Deletes from the map and stops there.
- `FastifyWebSocket.attach()`, inside the `socket.on('close', ...)` handler. Same as Real, deletes from the map and stops there.

In tests, the disconnect is observable. In production, it is not. There is no event going upward, no callback for `Publications` to register against. So in production today, when a user closes their tab, the watcher `Publications` opened for them stays alive on the Mongo emitter forever. Every change event still fires its callback. That is a real leak, and it is the dominant case in production because most clients close tabs without sending `unsubscribe`.

### What this points at

By the time you have done step 3, three things need fixing together:

1. A way for the test to ask "how many listeners are on this emitter right now". The `EventEmitter` in `outputTracker.ts` does not expose that today. A small method in the same family as the `off` you added gives you a real assertion. Listener count goes up on subscribe, back down on disconnect.

2. A way for the wrapper to surface disconnect to consumers. `WebSocketWrapper` has `trackDisconnections()` for tests but no callback API for `Publications`. The shape you want is the same as the `watchChanges(collection, cb)` you wrote in `mongo.ts`. A method that takes a callback and returns a cleanup function. `RealWebSocket` and `FastifyWebSocket` then need an internal emitter and need to fire it inside their `socket.on('close', ...)` handlers.

3. A wire up in `Publications`. Your `subscriptions` map is already keyed by `clientId`. Once the disconnect signal arrives, the teardown is one lookup and a loop over the cleanup functions you are already storing on subscribe.

### Scope of this PR

The leak gets fixed on this PR. Your `subscriptions` map and the `watchChanges` shape on `mongo.ts` are both already pointing at it, so the fold in is small. Today.

### Still open from last round

`isChangeEvent` only checks `type`, but the `ChangeEvent` union now says insert and update carry `fields`. Either tighten the guard at the boundary or relax the union, whichever feels more honest. Not blocking.

PR #32, opened 2026-04-27
Eko 227/3b1 handle added messages

Iam-Ebuka:
test: red - inserts a document on added message
test: green - inserts a document on added message
test: red - returns a single document by id
test: green - returns a single document by id
refactor: used helper function withId in reactiveStore

https://linear.app/ekohacks/issue/EKO-227/3b1-handle-added-messages

### DIFF
--- a/client/reactiveStore.ts
+++ b/client/reactiveStore.ts
@@ -22,12 +22,8 @@ export class ReactiveStore {
         break;
 
       case 'changed':
-        // handle changed message
-        break;
-
       case 'removed':
-        // handle removed message
-        break;
+        throw new Error(`Not implemented: ${msg.type}`);
 
       default:
         assertNever(msg);
--- a/client/reactiveStore.ts
+++ b/client/reactiveStore.ts
@@ -2,25 +2,44 @@ import { DataMsg } from '../shared/protocol.ts';
 
 type StoredDoc = Record<string, unknown>;
 
-function withId(id: string, fields: StoredDoc): StoredDoc & { _id: string } {
-  return { _id: id, ...fields };
+type StoredDocWithId = StoredDoc & { _id: string };
+
+function withId(id: string, fields: StoredDoc): StoredDocWithId {
+  return { ...fields, _id: id };
+}
+
+function assertNever(x: never): never {
+  throw new Error(`Unexpected value: ${JSON.stringify(x)}`);
 }
 
 export class ReactiveStore {
-  private serverMessages = new Map<string, Record<string, unknown>>();
+  private docs = new Map<string, Record<string, unknown>>();
 
   handleMessage(msg: DataMsg): void {
-    if (msg.type === 'added') {
-      this.serverMessages.set(msg.id, msg?.fields ?? {});
+    switch (msg.type) {
+      case 'added':
+        this.docs.set(msg.id, msg.fields ?? {});
+        break;
+
+      case 'changed':
+        // handle changed message
+        break;
+
+      case 'removed':
+        // handle removed message
+        break;
+
+      default:
+        assertNever(msg);
     }
   }
 
   getAll(): StoredDoc[] {
-    return Array.from(this.serverMessages.entries()).map(([id, fields]) => withId(id, fields));
+    return Array.from(this.docs.entries()).map(([id, fields]) => withId(id, fields));
   }
 
   getById(id: string): StoredDoc | undefined {
-    const fields = this.serverMessages.get(id);
+    const fields = this.docs.get(id);
     if (fields) {
       return withId(id, fields);
     }
--- a/client/reactiveStore.ts
+++ b/client/reactiveStore.ts
@@ -1,0 +1,18 @@
+import { DataMsg } from '../shared/protocol.ts';
+
+export class ReactiveStore {
+  private serverMessages = new Map<string, Record<string, unknown>>();
+
+  handleMessage(msg: DataMsg): void {
+    if (msg.type === 'added') {
+      this.serverMessages.set(msg.id, msg?.fields ?? {});
+    }
+  }
+
+  getAll(): Record<string, unknown>[] {
+    return Array.from(this.serverMessages.entries()).map(([id, fields]) => ({
+      _id: id,
+      ...fields,
+    }));
+  }
+}
--- a/client/reactiveStore.ts
+++ b/client/reactiveStore.ts
@@ -1,5 +1,11 @@
 import { DataMsg } from '../shared/protocol.ts';
 
+type StoredDoc = Record<string, unknown>;
+
+function withId(id: string, fields: StoredDoc): StoredDoc & { _id: string } {
+  return { _id: id, ...fields };
+}
+
 export class ReactiveStore {
   private serverMessages = new Map<string, Record<string, unknown>>();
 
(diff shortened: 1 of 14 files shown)


KevinOrfas:
## First pass on 3.B.1

A Few things I'd want to read with you rather than tell you straight.

`withId`, the spread. `{ _id: id, ...fields }`. Read it slowly and ask which `_id` wins when `fields` has one of its own. Spread order matters: whichever side comes second overrides the first. So if a server `added` message ever ships a `_id` inside `fields`, the parameter you carefully passed in is the one that loses. Whether that's right depends on what you decide the DDP envelope is meant to mean. My read: the envelope is the source of truth on identity, not the body. A test with `id: 'a'` and `fields: { _id: 'b' }` will tell you which way round you've actually wired it.

`handleMessage`, the line `this.serverMessages.set(msg.id, msg?.fields ?? {})`. Two guards stacked. Open `shared/protocol.ts` (or wherever `DataMsg` lives) and look at the `added` variant. The type is already telling you something. `msg` is non nullable here because it's a typed parameter, so the `?.` is doing nothing the compiler hasn't already done. The `?? {}` is the only one that could still earn its keep, and it earns it only if `fields` is actually optional on `added`. If `fields` is required on the type, the fallback is dead code. If it's optional, `?? {}` stays and `?.` goes. Worth checking which it is.

`serverMessages` as a name. Map holds fields, not messages. Reads slightly off today. Once 3.B.3 lands and `changed` and `removed` are also writing into the same map, "messages" starts to lie out loud. The map is a collection of documents keyed by id, so `documents` (or `docs`) reads truer. Cosmetic, but the kind of thing that's painless now and irritating in two stories.

The branch shape. `handleMessage` does something on `added` and silently returns on everything else. `DataMsg` grows in 3.B.3. Sit with the question of where you'd want a forgotten branch to surface: at build, in tests, or in production. The pattern that pays best here is a `switch` on `msg.type` with a `default` arm that does `const _: never = msg`. TypeScript refuses to compile the moment a new variant is added to the union without a matching case. You get the warning the second you touch `DataMsg` in 3.B.3, not the third time someone debugs why `removed` events aren't doing anything. The same question gets sharper once 3.B.2 lands, because then a missed branch doesn't just skip storage, it skips firing the change event too, which is a much quieter bug to chase.

`withId` is typed `StoredDoc & { _id: string }`. `getAll()` is typed `StoredDoc[]`. Put those two next to each other and ask what a consumer of `getAll()` sees on hover. They get back `Record<string, unknown>[]`, not the precise type the helper just guaranteed. The intersection type that `withId` produces gets discarded at the function boundary. Two ways out: change the return type of `getAll()` to `Array<StoredDoc & { _id: string }>`, or pull a named alias (`type StoredDocWithId = ...`) and use it on both. Either keeps the helper's promise visible to the caller.

Tests read clean. Indentation on the second `it` drifts, prettier handles it.

Of the five, the spread order is the only one that can produce a wrong answer at runtime today. The rest is the code asking you what kind of code it wants to be. Curious which one you reach for first.

KevinOrfas:
## Third pass

Most of it landed. Three things still want a look.

`withId`. Spread is the right way round now, envelope wins. Fix is in but nothing in the tests is holding it in place. A single assertion with `id: 'a'` and `fields: { _id: 'b' }` against `getById('a')` would lock the contract so a future refactor can't quietly flip the order back. Worth adding while you're here.

The switch with `assertNever`. Nice. That's the shape that catches a missing variant at compile time. Now read the two empty branches:

```ts
case 'changed':
  // handle changed message
  break;

case 'removed':
  // handle removed message
  break;
```

What does the safety net actually catch in 3.B.3? `assertNever` only fires when a new variant is added to `DataMsg`. The cases for `changed` and `removed` already exist, so if 3.B.3 ships and forgets to fill `'changed'` in, the compiler stays quiet and a `changed` message arriving at runtime does nothing. The empty branches let the work look done from the type system's point of view. Two ways to keep the net working. Delete the empty cases entirely, so `assertNever` catches them at runtime today and the compiler catches them in 3.B.3 the moment you put them back. Or fill them with `throw new Error('not yet handled')` so anything arriving early fails loudly. Either gives you the signal you set up the switch for in the first place.

`StoredDocWithId` alias. You pulled it out, good. It only lands on `withId`'s return type though. `getAll(): StoredDoc[]` and `getById(): StoredDoc | undefined` still hide the `_id` from callers. The alias is sitting right there waiting. Putting it on both public signatures means consumers see the contract the helper is already enforcing.

Tests. Indentation on the second `it` still drifts, prettier on save sorts it.

The spread test is the one I'd most want in this PR before merge. The other two are cheap. Curious which order you want to tackle them.

KevinOrfas:
## Quick follow up on the empty cases

Want to walk the switch question one step further, because reading my own comment back I was a bit indirect about what's actually at stake.

The `assertNever(msg)` in the default arm catches one thing well: a new variant added to `DataMsg` with no matching `case`. The compiler stops you. That's the whole win of an exhaustive switch.

What it cannot see is a `case` arm that exists but does nothing. The arm satisfies the union as far as the type system is concerned. The default never runs. The compiler is happy.

So picture 3.B.3. You sit down to handle `changed` and `removed`. Empty stubs are already there. You write the `changed` body, run the tests, they go green. You move on. The `removed` stub is still empty. Nothing in the build, nothing in the type checker, nothing in the existing tests is going to tell you. A `removed` message arrives at runtime, the store keeps the document, a view shows a ghost. You only find that one in code review or by writing the test that catches it before you've written the code.

Likely you would catch it. Story's literally called "handle changed and removed", you'd see both stubs. But the design today asks you to remember. The exhaustive switch was supposed to take that job off your plate.

The cheap move: delete the two empty cases. `assertNever(msg)` then does two jobs at once. Right now, if anything ever did send a `changed` or `removed` (server doesn't yet, but if it did), the function throws loudly instead of silently swallowing. And in 3.B.3, the moment you reach for `'changed'` or `'removed'` you'll be adding the case back with a real body, because the compiler will refuse to build without it.

Tiny code change, makes the safety net actually catch what you set it up to catch.

KevinOrfas:
## Fourth pass

The empty case fix is exactly the shape. `case 'changed': case 'removed': throw`. Falls through to one body, anything arriving at runtime fails loud, 3.B.3 will trip the moment it tries to make a real `changed` or `removed` test pass. Clean.

Two from the third pass still open.

Spread test. The fix is in the code (`{ ...fields, _id: id }`) but the test suite doesn't know about it. Right now if someone flipped that line back the way it was, nothing in `reactiveStore.test.ts` would catch it. A single assertion locks it in:

```ts
store.handleMessage({
  type: 'added',
  collection: 'files',
  id: 'envelope-id',
  fields: { _id: 'body-id', name: 'thing.bam' },
});
expect(store.getById('envelope-id')).toEqual({ _id: 'envelope-id', name: 'thing.bam' });
```

If the spread were the wrong way round, that assertion fails. That's the contract you decided when you flipped the order, written down in a way the next person can't quietly undo.

`StoredDocWithId` at the boundary. You pulled the alias out, applied it to `withId`'s return. `getAll(): StoredDoc[]` and `getById(): StoredDoc | undefined` are still typed in terms of the looser `StoredDoc`. The helper guarantees the `_id`, the public API hides it. Switch the two return signatures over and consumers see the contract you've actually got.

Indentation on the second `it` still drifts. Prettier handles it on save.

Both small. The spread test is the one that matters for protecting the runtime fix you just made.

KevinOrfas (approved):

PR #33, opened 2026-04-27
Eko 226/3b2 emit change events

Iam-Ebuka:
- test: red - calls registered listeners when a document is added
- test: green - calls registered listeners when a document is added
- test: red - stops calling the listener after off() is called
- test: green - stops calling the listener after off() is called

https://linear.app/ekohacks/issue/EKO-226/3b2-emit-change-events

### DIFF
--- a/client/reactiveStore.ts
+++ b/client/reactiveStore.ts
@@ -21,7 +21,7 @@ export class ReactiveStore {
     switch (msg.type) {
       case 'added':
         this.docs.set(msg.id, msg.fields ?? {});
-        this.emitter.emit('change', null);
+        this.emitter.emit('change');
         break;
 
       case 'changed':
--- a/client/reactiveStore.ts
+++ b/client/reactiveStore.ts
@@ -22,12 +22,8 @@ export class ReactiveStore {
         break;
 
       case 'changed':
-        // handle changed message
-        break;
-
       case 'removed':
-        // handle removed message
-        break;
+        throw new Error(`Not implemented: ${msg.type}`);
 
       default:
         assertNever(msg);
--- a/client/reactiveStore.ts
+++ b/client/reactiveStore.ts
@@ -2,25 +2,44 @@ import { DataMsg } from '../shared/protocol.ts';
 
 type StoredDoc = Record<string, unknown>;
 
-function withId(id: string, fields: StoredDoc): StoredDoc & { _id: string } {
-  return { _id: id, ...fields };
+type StoredDocWithId = StoredDoc & { _id: string };
+
+function withId(id: string, fields: StoredDoc): StoredDocWithId {
+  return { ...fields, _id: id };
+}
+
+function assertNever(x: never): never {
+  throw new Error(`Unexpected value: ${JSON.stringify(x)}`);
 }
 
 export class ReactiveStore {
-  private serverMessages = new Map<string, Record<string, unknown>>();
+  private docs = new Map<string, Record<string, unknown>>();
 
   handleMessage(msg: DataMsg): void {
-    if (msg.type === 'added') {
-      this.serverMessages.set(msg.id, msg?.fields ?? {});
+    switch (msg.type) {
+      case 'added':
+        this.docs.set(msg.id, msg.fields ?? {});
+        break;
+
+      case 'changed':
+        // handle changed message
+        break;
+
+      case 'removed':
+        // handle removed message
+        break;
+
+      default:
+        assertNever(msg);
     }
   }
 
   getAll(): StoredDoc[] {
-    return Array.from(this.serverMessages.entries()).map(([id, fields]) => withId(id, fields));
+    return Array.from(this.docs.entries()).map(([id, fields]) => withId(id, fields));
   }
 
   getById(id: string): StoredDoc | undefined {
-    const fields = this.serverMessages.get(id);
+    const fields = this.docs.get(id);
     if (fields) {
       return withId(id, fields);
     }
--- a/client/reactiveStore.ts
+++ b/client/reactiveStore.ts
@@ -33,11 +33,11 @@ export class ReactiveStore {
     }
   }
 
-  getAll(): StoredDoc[] {
+  getAll(): StoredDocWithId[] {
     return Array.from(this.docs.entries()).map(([id, fields]) => withId(id, fields));
   }
 
-  getById(id: string): StoredDoc | undefined {
+  getById(id: string): StoredDocWithId | undefined {
     const fields = this.docs.get(id);
     if (fields) {
       return withId(id, fields);
--- a/client/reactiveStore.ts
+++ b/client/reactiveStore.ts
@@ -30,7 +30,11 @@ export class ReactiveStore {
     return undefined;
   }
 
-  onChange(listener: () => void): void {
+  onChange(listener: () => void): () => void {
     this.emitter.on('change', listener);
+
+    return () => {
+      this.emitter.off('change', listener);
+    };
   }
 }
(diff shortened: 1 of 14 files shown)


KevinOrfas (commented):
## Reading the diff

- The red-green-red-green commit shape made it easy to follow what each step was for, and to see that 3.B.1's tests still pass on top.
- Listener signature stays no-arg, so the protocol shape doesn't leak into the UI contract.
- Emit fires *after* `this.docs.set(...)`. That ordering matters: if emit ran first, listeners would call `getAll()` and read the *old* state because the Map hasn't been updated yet. Mutate first, emit second is the safe order, and that's how it lands here.
- Reusing the existing EventEmitter means `outputTracker.ts` didn't need touching, which keeps the diff tight to the story.
- The returned `off()` closes over the same `listener` reference passed into `emitter.on`. That's the bit that makes unregister actually work, because the EventEmitter matches by reference.

## Things to tighten before merge

### The `null` payload

```ts
this.emitter.emit('change', null);
```

The listener takes no arguments, so `null` is along for the ride. The shape is forced by the EventEmitter's signature `emit(eventType, data: unknown)`. Two ways out:

- Make `data` optional on `EventEmitter.emit` so callers can write `this.emitter.emit('change')`. Small change, helps every future caller.
- Leave it, treat the `null` as the EventEmitter's quirk to fix another day.

What feels right to you? If you do touch the emitter, keep it to that one parameter so the diff matches the story title.

### Coverage gaps worth ironing in this PR

Three behaviours aren't covered by a test yet. Better to lock them down here than carry them into 3.B.3.

**1. Multiple listeners both fire on the same change.**

Quick orientation on what that means. Inside `EventEmitter` (`server/infrastructure/outputTracker.ts`):

```ts
private handlers: Map<string, ((data: unknown) => void)[]> = new Map();
```

`handlers` is the *registry*, the place where listeners are stored. It maps an event name like `'change'` to an *array of listener functions*. When you call `onChange(fn)`, that `fn` gets pushed onto the array under `'change'`. When `emit('change', ...)` runs, the code loops the array and calls each function in turn.

Today, two listeners both firing works because the storage shape is an array. No test asserts it. Picture a future refactor:

- Someone rewrites `handlers` as `Map<string, listener>` thinking 'we only ever have one listener per event.' A second `onChange` call now silently overwrites the first.
- Someone swaps the array for a `Set` thinking 'Sets are tidier.' Sets dedupe by reference, so passing the same function twice keeps only one copy, and listener counts stop matching `onChange` call counts.

Either change passes the existing tests and quietly breaks the contract. A test that registers two listeners and asserts both got called turns the implicit guarantee into an explicit one.

**2. `off()` called twice doesn't crash.**

Today it doesn't, because `indexOf` returns -1 the second time and the splice is guarded. A one-liner test keeps that guarantee load-bearing rather than incidental. Without the test, a future tidy of `off` could remove the guard and nothing would notice.

**3. Listener fires per mutation, not per read.**

A test that calls `getAll()` between mutations and watches the count stay flat would make this explicit. Cheap to add, makes the contract obvious to the next reader.

### Latent double-fire bug when the same listener is registered twice

This one is worth fixing in this PR, not flagging for later. Here's the shape:

```ts
const fn = () => console.log('hi');
const off1 = store.onChange(fn);
const off2 = store.onChange(fn);

store.handleMessage({ type: 'added', ... });
// 'hi' prints TWICE because fn is in the array twice

off1();
// fn is still in the array, one copy remains

store.handleMessage({ type: 'added', ... });
// 'hi' still prints, even though we thought we cleaned up
```

`EventEmitter.on` pushes onto the array unconditionally, so the same function reference can sit in there many times. `off()` uses `indexOf`, which finds the first match and removes only one copy. The visible symptoms when this bites: duplicate re-renders, listeners that won't go away after teardown.

It hasn't bitten yet because nothing in the codebase double-registers. But the moment a real UI consumer wires this up it can, especially in patterns where a `useEffect` runs twice on mount or a component re-renders before its cleanup ran. That's the kind of bug you only catch in production when something feels off and you spend an afternoon tracing why a handler runs four times.

Two clean ways to close it:

1. *Dedupe at registration in `EventEmitter.on`.* Before pushing, check if the handler is already in the array. If yes, no-op. Smallest change, and it sets the principle that 'registering twice has the same effect as registering once' across every caller of this class.

2. *Wrap in `ReactiveStore.onChange`.* The store keeps a `Set<listener>` of its own. Adding a listener that's already in the Set is a no-op, and the `off()` returned is the same regardless of how many `onChange(fn)` calls happened. Keeps the emitter alone, contains the change to this story.

My lean is option 1. The fix is small, the test is small, and it removes the bug class from every other caller of `EventEmitter` rather than just this one. Curious how you read it.

## Memory leak shape to keep in mind

The risk this PR carries doesn't live in the PR itself, it lives in whoever uses `onChange`. If a consumer registers a listener and never calls the returned `off()`, the listener stays in the array forever. Every mount-unmount cycle stacks another dead listener and the leak grows. We hit the same shape at the socket layer in 3.A.4 (PR #31, the disconnect-leak fix), where the integration test asserts that the listener count drops to zero after disconnect. That same test pattern is the one to reach for when `onChange` gets wired into a real component.

KevinOrfas:
## Follow-up: memory leak shape and where the fix lives

Splitting this out from the main review because it's less about the diff and more about the next step that this `onChange` API opens up. Worth reading once the rest of the review is in.

### Why `ReactiveStore` can't fix the leak on its own

The store has no signal for 'the consumer is done.' React components mount and unmount. Svelte stores get destroyed. Vanilla JS callers lose their reference. None of that is visible from inside `ReactiveStore`, so the store cannot unilaterally clean up on the consumer's behalf. The discipline has to live where the consumer's lifecycle lives.

Same constraint as 3.A.4. The publication manager couldn't clean itself up either, the socket disconnect event was the lifecycle hook that did it. Same shape here, different layer.

### Where the fix actually lives

Two places, in this order.

**1. In the consumer's lifecycle binding.**

When the first real consumer of `onChange` ships (a React hook, a Svelte store wrapper, whatever), it pairs `onChange` with the consumer's teardown signal:

```ts
// React
useEffect(() => store.onChange(rerender), [store]);
// The function returned by onChange is what React calls on unmount.

// Svelte
onDestroy(store.onChange(rerender));

// Vanilla
const off = store.onChange(rerender);
element.addEventListener('disconnect', off);
```

The pattern: the framework already has a teardown event, and `onChange`'s return value plugs straight into it. If you forget to return it, the framework can't clean up, and the leak surfaces in tests.

**2. In the consumer's tests.**

The proof is the same one PR #31 used at the socket layer: assert listener count drops to zero after teardown.

```ts
it('removes its listener on unmount', () => {
  const store = new ReactiveStore();
  const { unmount } = render(<MyComponent store={store} />);

  expect(store.listenerCount()).toBe(1);

  unmount();

  expect(store.listenerCount()).toBe(0);
});
```

That test stays red until the cleanup is wired correctly. Exact shape as the disconnect-leak test from PR #31, translated to the mount lifecycle.

### One small move that would help the future test

`ReactiveStore` could expose `listenerCount()` as a thin pass-through to `this.emitter.listenerCount('change')`. Three lines, zero behaviour change for current callers, and it means the consumer's leak test doesn't have to reach into the emitter directly.

Optional. If you'd rather keep this PR tight to the story, the test can read off the emitter directly when the consumer lands. Both work. Curious which feels right.

KevinOrfas (approved):

KevinOrfas (approved):
All review points addressed. Approving. Posted a small follow-up tidy at #34, stacked on this branch — merge that after this one lands on main.

PR #34, opened 2026-05-04
refactor: tidy EventEmitter.on so the array reference is always the stored one

KevinOrfas:
## Summary
- Stacked on #33 (the dedupe lands there). This is the small follow-up tidy we agreed in the review thread.
- `EventEmitter.on` previously did `this.handlers.get(eventType) ?? []` after a `has`-check. The `?? []` fallback is dead at runtime, but if `get` ever did return `undefined`, `handlers` would be a fresh array that was never stored back on the Map and the push would silently vanish.
- Folded the two steps so the local always references the array that's actually stored. Same behaviour, same listenerCount semantics.

## Why
The `?? []` was a TypeScript shape concession, not a runtime guard. Removing it drops a dead path and makes the relationship between the local and the Map obvious. Surfaced while reviewing #33.

## Test plan
- [x] `npx vitest run` (full suite green)
- [x] `npm run lint` (no new warnings)
- [x] `npx tsc --noEmit` (clean)
- [ ] Merge after #33 lands on main; the diff on this PR will collapse to just the one commit at that point.

### DIFF
--- a/client/reactiveStore.ts
+++ b/client/reactiveStore.ts
@@ -21,7 +21,7 @@ export class ReactiveStore {
     switch (msg.type) {
       case 'added':
         this.docs.set(msg.id, msg.fields ?? {});
-        this.emitter.emit('change', null);
+        this.emitter.emit('change');
         break;
 
       case 'changed':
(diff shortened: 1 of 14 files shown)


Iam-Ebuka (approved):
I see, nice one. Resolve conflict needed

PR #35, opened 2026-05-04
Eko 225/3b3 handle changed and removed

Iam-Ebuka:
test: red - merges fields into the existing document on changed message
test: green - merges fields into the existing document on changed mes…
test: red - deletes the document on removed message
test: green - deletes the document on removed message
refactor: emit change should live in one place
test: green - emits a change event for added, changed, and removed

https://linear.app/ekohacks/issue/EKO-225/3b3-handle-changed-and-removed

### DIFF
--- a/client/reactiveStore.ts
+++ b/client/reactiveStore.ts
@@ -21,10 +21,20 @@ export class ReactiveStore {
     switch (msg.type) {
       case 'added':
         this.docs.set(msg.id, msg.fields ?? {});
-        this.emitter.emit('change', null);
+        this.emitter.emit('change');
         break;
 
-      case 'changed':
+      case 'changed': {
+        const existing = this.docs.get(msg.id);
+        if (!existing) {
+          return;
+        }
+
+        this.docs.set(msg.id, { ...existing, ...msg.fields });
+        this.emitter.emit('change');
+        break;
+      }
+
       case 'removed':
         throw new Error(`Not implemented: ${msg.type}`);
 
--- a/client/reactiveStore.ts
+++ b/client/reactiveStore.ts
@@ -22,12 +22,8 @@ export class ReactiveStore {
         break;
 
       case 'changed':
-        // handle changed message
-        break;
-
       case 'removed':
-        // handle removed message
-        break;
+        throw new Error(`Not implemented: ${msg.type}`);
 
       default:
         assertNever(msg);
--- a/client/reactiveStore.ts
+++ b/client/reactiveStore.ts
@@ -2,25 +2,44 @@ import { DataMsg } from '../shared/protocol.ts';
 
 type StoredDoc = Record<string, unknown>;
 
-function withId(id: string, fields: StoredDoc): StoredDoc & { _id: string } {
-  return { _id: id, ...fields };
+type StoredDocWithId = StoredDoc & { _id: string };
+
+function withId(id: string, fields: StoredDoc): StoredDocWithId {
+  return { ...fields, _id: id };
+}
+
+function assertNever(x: never): never {
+  throw new Error(`Unexpected value: ${JSON.stringify(x)}`);
 }
 
 export class ReactiveStore {
-  private serverMessages = new Map<string, Record<string, unknown>>();
+  private docs = new Map<string, Record<string, unknown>>();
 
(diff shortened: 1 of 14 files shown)


KevinOrfas (commented):
## What this lands
The other two arms of the message handler. `changed` merges fields into the existing doc, `removed` deletes by id, and the emit moves to a single call at the bottom of `handleMessage` so all three branches notify listeners in the same place.

## Reading the diff
- The single emit after the switch is a nice refactor. Means `added`, `changed`, and `removed` all notify listeners through the same code path. Removes the duplication that would have crept in if each branch had its own `emit` call. The commit "refactor: emit change should live in one place" earned its name.
- Shallow merge with `{ ...existing, ...msg.fields }` is the right call for this protocol shape. The test pins it down with the `name` key surviving while `size` gets updated.
- `Map.delete` for `removed` is exactly the simple thing you'd expect.
- Tests for the three behaviours are present and small. The "emits a change event for added, changed, and removed" test is the strongest of the three because it wires the emit-once-from-the-bottom move to the contract.

## One thing worth a chew

There's an asymmetry between `changed` and `removed` when the id is unknown to the store.

```ts
case 'changed': {
  const existing = this.docs.get(msg.id);
  if (!existing) {
    return;
  }
  this.docs.set(msg.id, { ...existing, ...msg.fields });
  break;
}

case 'removed':
  this.docs.delete(msg.id);
  break;
```

`changed` for an unknown id short circuits with `return`, so it skips the emit at the bottom. `removed` for an unknown id just calls `Map.delete` (which silently does nothing for missing keys) and then falls through to the emit. So:

- `changed` on unknown id: no state change, no emit.
- `removed` on unknown id: no state change, but emit still fires.

Two questions land in here.

First, is the silent skip on `changed` what you actually want? In a sync protocol, a `changed` for an id the client doesn't have means a gap between server and client state. Possible reads of that gap:

- *Silent skip* (current). Cheap, the client carries on, but it stays out of sync until the next added or a resync.
- *Throw or log.* Signals the protocol is in a state it shouldn't be in. Useful for debugging during the bootcamp, less so for production.
- *Treat it as added.* You have fields, store them. Risky if `msg.fields` is partial in a way `added` wouldn't be.

Silent skip is defensible. Worth being explicit that you chose it, because the next person reading this will wonder.

Second, the asymmetric emit. If the rule is "every successful mutation notifies listeners," then `removed` on unknown id shouldn't emit either. Either short circuit both, or neither.

```ts
case 'removed':
  if (!this.docs.has(msg.id)) {
    return;
  }
  this.docs.delete(msg.id);
  break;
```

Five extra lines, and the contract becomes "one emit per real mutation, full stop." Or go the other way and let `changed` fall through with a no op set when the id is unknown. Either is consistent with itself. The current shape mixes them.

## Coverage gap worth a one liner

There's no test for the unknown id paths in either branch. Two tiny tests would lock down whichever behaviour you settle on:

```ts
it('does nothing when changed arrives for an unknown id', () => {
  const store = new ReactiveStore();
  let callCount = 0;
  store.onChange(() => callCount++);

  store.handleMessage({ type: 'changed', collection: 'files', id: 'ghost', fields: { name: 'a' } });

  expect(store.getById('ghost')).toBeUndefined();
  expect(callCount).toBe(0);
});

it('does nothing when removed arrives for an unknown id', () => {
  const store = new ReactiveStore();
  let callCount = 0;
  store.onChange(() => callCount++);

  store.handleMessage({ type: 'removed', collection: 'files', id: 'ghost' });

  expect(store.getById('ghost')).toBeUndefined();
  expect(callCount).toBe(0);
});
```

The second one fails on the current implementation because `removed` emits even when nothing was deleted. That failing test is the reason to make the two branches symmetric.

## Risk read

Additive only on the message handler. `added` behaviour is unchanged (the move from inline emit to bottom emit is functionally identical for that branch). `changed` and `removed` previously threw, so anything subscribing to live data on the client now actually works instead of crashing. Nothing in the existing 3.B.1 or 3.B.2 tests changes shape. Story 3.A tests on the server side don't touch this code.

The only behavioural risk lives in the asymmetric emit on unknown id, which is what the chew above is about.

KevinOrfas (approved):

KevinOrfas (approved):

Issue #2, opened 2026-04-02
Install git commit package


KevinOrfas:
As a dev, I want to be able to understand the commits we write and establish conventions across the team

PR #38, opened 2026-05-05
Eko 271/3d observer interface for reactivestore and publications

Iam-Ebuka:
- refactor: added observerOutcome and reactiveStoreObserver interface
- refactor: added constructor to accomodate reactiveStoreObserver in reactorStore.ts
- test: green - notifies observer on applied added message
- test: green - notifies observer when changed arrives for an unknown id
- test: green - ignores observer errors and continues normal processing
- refactor: reactiveStoreObserver
- test: red - notifies observer on failed subscribe for unknown publication
- test: green - notifies observer on failed subscribe for unknown publication
- test: green - notifies observer on applied subscribe request
- test: green - notifies observer when unsubscribe cannot find the sub id
- test: green - continues normal behavior if observer throws

- refactor: watching out for unrelated event that might slip in
- test: green - replaces the watcher when the same client resubscribes with the same id
- refactor: added reasons code documentation for observers in protocols.ts
- test: green - notifies observer when removed message arrives for unknown id
- refactor: seperated interface so we know have reactStoreObserver and publicationsObserver

https://linear.app/ekohacks/issue/EKO-271/3d-observer-interface-for-reactivestore-and-publications

### DIFF
--- a/client/reactiveStore.ts
+++ b/client/reactiveStore.ts
@@ -1,5 +1,5 @@
 import { EventEmitter } from '../server/infrastructure/outputTracker.ts';
-import { DataMsg } from '../shared/protocol.ts';
+import { DataMsg, ReactiveStoreObserver } from '../shared/protocol.ts';
 
 type StoredDoc = Record<string, unknown>;
 
@@ -16,31 +16,54 @@ function assertNever(x: never): never {
 export class ReactiveStore {
   private docs = new Map<string, Record<string, unknown>>();
   private emitter = new EventEmitter();
+  private observer: ReactiveStoreObserver;
+
+  constructor(observer: ReactiveStoreObserver = { onMessage: () => {} }) {
+    this.observer = observer;
+  }
+
+  private notifyObserver(
+    msg: DataMsg,
+    outcome: 'applied' | 'skipped' | 'failed',
+    reason?: string,
+  ): void {
+    try {
+      this.observer.onMessage(msg, outcome, reason);
+    } catch (err) {
+      console.error('Observer error:', err);
+    }
+  }
 
   handleMessage(msg: DataMsg): void {
     switch (msg.type) {
       case 'added':
         this.docs.set(msg.id, msg.fields ?? {});
+        this.notifyObserver(msg, 'applied');
         break;
 
       case 'changed': {
         const existing = this.docs.get(msg.id);
         if (!existing) {
+          this.notifyObserver(msg, 'skipped', 'unknown-id');
           return;
         }
 
         this.docs.set(msg.id, { ...existing, ...msg.fields });
+        this.notifyObserver(msg, 'applied');
         break;
       }
 
       case 'removed':
         if (!this.docs.has(msg.id)) {
+          this.notifyObserver(msg, 'skipped', 'unknown-id');
           return;
         }
         this.docs.delete(msg.id);
+        this.notifyObserver(msg, 'applied');
         break;
 
       default:
+        this.notifyObserver(msg, 'failed', 'unsupported-message-type');
         assertNever(msg);
     }
     this.emitter.emit('change');
--- a/client/reactiveStore.ts
+++ b/client/reactiveStore.ts
@@ -1,5 +1,5 @@
 import { EventEmitter } from '../server/infrastructure/outputTracker.ts';
-import { DataMsg, ReactiveStoreObserver } from '../shared/protocol.ts';
+import { DataMsg, ReactiveStoreObserver, ReactiveStoreReasons } from '../shared/protocol.ts';
 
 type StoredDoc = Record<string, unknown>;
 
@@ -25,7 +25,7 @@ export class ReactiveStore {
   private notifyObserver(
     msg: DataMsg,
     outcome: 'applied' | 'skipped' | 'failed',
-    reason?: string,
+    reason?: ReactiveStoreReasons,
   ): void {
     try {
       this.observer.onMessage(msg, outcome, reason);
--- a/server/logic/publications.ts
+++ b/server/logic/publications.ts
@@ -1,4 +1,4 @@
-import { ClientMessage } from '../../shared/protocol.ts';
+import { ClientMessage, ReactiveStoreObserver } from '../../shared/protocol.ts';
 import { MongoWrapper } from '../infrastructure/mongo.ts';
 import { WebSocketWrapper } from '../infrastructure/websocket.ts';
 import { ChangeEvent } from '../../shared/types.ts';
@@ -22,13 +22,31 @@ export class Publications {
   private ws: WebSocketWrapper;
   private mongo: MongoWrapper;
   private subscriptions = new Map<string, Map<string, () => void>>();
+  private observer: ReactiveStoreObserver;
 
-  constructor(mongo: MongoWrapper, ws: WebSocketWrapper) {
+  constructor(
+    mongo: MongoWrapper,
+    ws: WebSocketWrapper,
+    observer: ReactiveStoreObserver = { onMessage: () => {} },
+  ) {
     this.mongo = mongo;
     this.ws = ws;
+    this.observer = observer;
     this.ws.onDisconnect((clientId) => {
       this.tearDownClient(clientId);
     });
+  }
+
+  private notifyObserver(
+    msg: ClientMessage,
+    outcome: 'applied' | 'skipped' | 'failed',
+    reason?: string,
+  ): void {
+    try {
+      this.observer.onMessage(msg, outcome, reason);
+    } catch (err) {
+      console.error('Observer error:', err);
+    }
   }
 
   private tearDownClient(clientId: string): void {
@@ -56,6 +74,7 @@ export class Publications {
           id: message.id,
           error: { code: 404, message: `Unknown publication: ${message.name}` },
         });
+        this.notifyObserver(message, 'failed', 'unknown-publication');
         return Promise.resolve();
       }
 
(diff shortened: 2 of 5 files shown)


client/reactiveStore.ts
```diff
@@ -16,31 +16,54 @@ function assertNever(x: never): never {
...
+    } catch (err) {
+      console.error('Observer error:', err);
+    }
+  }
 
   handleMessage(msg: DataMsg): void {
     switch (msg.type) {
       case 'added':
         this.docs.set(msg.id, msg.fields ?? {});
+        this.notifyObserver(msg, 'applied');
```
KevinOrfas, line 41: The observer notify fires before `this.emitter.emit('change')` on line 69. So an external observer sees `'applied'` before the public `'change'` listener fires. Probably what you want, just flagging. Once a Sentry or OTel hook lands, that ordering becomes part of the observable contract, so worth being deliberate about it now.

client/reactiveStore.ts
```diff
@@ -16,31 +16,54 @@ function assertNever(x: never): never {
...
         if (!this.docs.has(msg.id)) {
+          this.notifyObserver(msg, 'skipped', 'unknown-id');
           return;
         }
         this.docs.delete(msg.id);
+        this.notifyObserver(msg, 'applied');
         break;
 
       default:
+        this.notifyObserver(msg, 'failed', 'unsupported-message-type');
```
KevinOrfas, line 66: We notify the observer with `'failed' / 'unsupported-message-type'`, then `assertNever(msg)` throws. Net behaviour: observer records the failure, process crashes, change event never fires. Probably right (the branch is unreachable in well typed code, the throw is a developer aid), but worth a one-line comment so a future reader doesn't read the notify-then-throw as an oversight.

server/logic/publications.ts
```diff
@@ -86,20 +105,29 @@ export class Publications {
       }
 
       clientSubs.set(message.id, cleanup);
```
KevinOrfas, line 107: The observer reports `'applied'` with reason `'duplicate-sub-id'` when a client reuses an existing sub id. The silent teardown of the prior watcher at lines 102 to 105 isn't pinned by any test. Since we're now surfacing the duplicate via the observer, let's add a failing test that pins both halves together: a second subscribe with the same id replaces the first watcher *and* the observer records the duplicate flag. Without that, the reason string is asserted in isolation from the behaviour it reports on.

server/logic/publications.ts
```diff
@@ -22,15 +22,33 @@ export class Publications {
...
   private mongo: MongoWrapper;
   private subscriptions = new Map<string, Map<string, () => void>>();
+  private observer: ReactiveStoreObserver;
 
-  constructor(mongo: MongoWrapper, ws: WebSocketWrapper) {
+  constructor(
+    mongo: MongoWrapper,
+    ws: WebSocketWrapper,
+    observer: ReactiveStoreObserver = { onMessage: () => {} },
+  ) {
```
KevinOrfas, line 31: Tiny one for the next commit: the message body on `083766d` says "accomodate" rather than "accommodate". Cosmetic, just so it doesn't repeat.

tests/client/reactiveStore.test.ts
```diff
@@ -15,6 +16,63 @@ describe('ReactiveStore', () => {
     expect(store.getAll()).toEqual([{ _id: '1', name: 'existing.bam' }]);
   });
 
+  it('notifies observer on applied added message', () => {
+    const applied: Array<{ outcome: string; reason?: string }> = [];
+    const store = new ReactiveStore({
+      onMessage(msg) {
```
KevinOrfas: The `if (msg.type === 'added')` filter inside the observer scopes the assertion neatly but lets any unrelated event slip past silently. If the intent is "exactly one applied event for this added message", capturing the whole sequence and asserting on the full list catches an unexpected extra notification too. Minor.

While you're in here: project style is `T[]` over `Array<T>` for collection types. Two spots in this file pick up `Array<...>` (line 20 and line 39).

KevinOrfas (changes requested):
Hi Ebuka, thanks for landing 3.D, the observer is exactly the shape the ticket calls out. Three things to sort before we ship.

**1. `vi.fn()` is the first mock to land. Let's stay consistent.**

`tests/logic/publications.test.ts` reaches for `vi.fn()` in four places (lines 26, 51, 92, 119). Our test convention is **Nullable Infrastructure** (James Shore): production objects expose `createNull()` to swap real I/O for in-memory equivalents, every other line of logic stays real, and we assert via an `OutputTracker` or a small recording double that captures what came out. We don't use `vi.fn`, `jest.fn`, sinon, or interaction-style spies. The reason: spies assert *the code under test made a particular call*, which drifts the moment the implementation refactors. Recorders assert *the right thing actually happened*, which survives refactors.

The good news, you've already used the right pattern on the client side (`tests/client/reactiveStore.test.ts:38`). Same shape on the server side and we're done:

```ts
const observed: { outcome: ObserverOutcome; reason?: string }[] = [];
const observer = {
  onMessage(msg, outcome, reason) {
    observed.push({ outcome, reason });
  },
};
const pubs = new Publications(mongo, ws, observer);
// ...
expect(observed).toContainEqual({ outcome: 'failed', reason: 'unknown-publication' });
```

No spy, no `vi`, just a recording double feeding an array. Keeps the suite uniform and stops the mocking surface from leaking further.

**2. DoD item: reason codes aren't documented.**

The ticket says "Reason codes documented for each skipped and failed path." Today the reason strings (`'unknown-id'`, `'unknown-publication'`, `'unknown-sub-id'`, `'duplicate-sub-id'`, `'unsupported-message-type'`) live as string literals scattered across the two implementation files. A consumer wiring up Sentry or OTel can't discover them without grepping the framework, and TypeScript doesn't catch a typo when they pattern match on `reason`.

Smallest move that closes the DoD: named string unions in `shared/protocol.ts`, then narrow the observer signature. Roughly:

```ts
export type ReactiveStoreSkipReason = 'unknown-id';
export type ReactiveStoreFailReason = 'unsupported-message-type';
export type PublicationsSkipReason = 'unknown-sub-id';
export type PublicationsFailReason = 'unknown-publication';
// duplicate-sub-id is an applied-with-reason on the subscribe path
```

That's the contract a consumer programmes against. A doc comment on each union explains what the path means and what a consumer should do with it.

**3. DoD item: missing `removed` observer path.**

The DoD says "Calls land at every message lifecycle event." The PR pins `added` to `applied` and `changed` against unknown id to `skipped` / `unknown-id`. But not the symmetric `removed` against unknown id, which `client/reactiveStore.ts:58` already notifies. One more failing test pinning the `'skipped' / 'unknown-id'` shape on the `removed` arm closes the gap.

Smaller bits inline. One side question: the ticket sketches `onMessage(msg: DataMsg, …)` for ReactiveStore specifically and says "Same interface on Publications." The PR landed one shared TS interface carrying `DataMsg | ClientMessage`. Did you read "same" as one widened interface, or as parallel-but-separate? Not blocking on its own, but it ties into #2 because reason codes split cleanly along that line, and the test at `tests/client/reactiveStore.test.ts:22` already shows the cost of widening (the consumer has to discriminate inside `onMessage`).

client/reactiveStore.ts
Iam-Ebuka: sure, worth being deliberate about it

server/logic/publications.ts
Iam-Ebuka: good catch

tests/client/reactiveStore.test.ts
Iam-Ebuka: instead of doing a conditional statement I would just check for the result and assert it is what I want it to be, that way we can catch any unrelated event

KevinOrfas (changes requested):
Most of this landed. `vi.fn()` is out, recording doubles read consistent on both sides, the duplicate-sub-id test pins watcher count and reason code together, the `removed`/unknown id arm has coverage. Splitting into two observer interfaces was your call and it reads better than the widened version I'd sketched.

Two threads still open.

**Documented means a comment per reason literal.**

The DoD reads "Reason codes documented for each skipped and failed path." You closed the typed half: `shared/protocol.ts:89` to `:99` carries named unions for every reason. The other half is a one line comment per literal saying when it fires and what a consumer should do with it. Concretely:

```ts
export type PublicationsReasons =
  /** Subscribe arrived with an id that already has a watcher on this
   *  client. Old watcher torn down, new one installed. Applied,
   *  not failed. */
  | 'duplicate-sub-id'
  /** Unsubscribe arrived for a sub id this client doesn't have.
   *  Usually a client bug or a race between client and server. */
  | 'unknown-sub-id'
  /** Subscribe arrived for a publication name that wasn't define()d
   *  on the server. Client and server schemas are out of sync. */
  | 'unknown-publication';
```

Same shape for `ReactiveStoreReasons`. The literal name says what happened, the comment says what it means. The consumer wiring a Sentry hook to `onMessage` reads the comment to make a paging decision without having to grep the framework.

**`Array<T>` reintroduction at `tests/client/reactiveStore.test.ts:58`.**

You fixed `Array<T>` to `T[]` on lines 20 and 39 last round. The new `removed` test at line 58 brought a third one back, almost certainly copied from a version of the file before you'd run the fix.

The fix is one keystroke. The pattern worth naming is: once you've normalised something across a file, anything you copy in afterwards needs the same pass, otherwise the diff fixes two and breaks one and the file ends up in a worse state than if you'd left it alone.

KevinOrfas (approved):

PR #1, opened 2026-04-01
Setting Up CI/CD

Okpala76:
- Installs an Ubuntu machine
-  makes a copy of our code
- installs Node.js version 22
- install repo dependency
- runs lint and 
- runs test

issue URL: https://linear.app/ekohacks/issue/EKO-188/story-cicd-and-pre-commit-hooks

PR #39, opened 2026-05-06
Eko 223/3c1 subscribe sends message and routes responses

Iam-Ebuka:
- test: red - subscribe sends a subscribe message and routes added documents into the store
- test: green - subscribe sends a subscribe message and routes added documents into the store
- introduced connectionManager class
- test: red - handle.stop sends an unsubscribe and stops further messages reaching the store
- test: green - handle.stop sends an unsubscribe and stops further messages reaching the store
- refactor: connectionManager.ts
- created a case to catch error in handleServerMessage
- refactored the test in clientWebSocket.test because server messages doesn't pass through trackMessages()

https://linear.app/ekohacks/issue/EKO-223/3c1-subscribe-sends-message-and-routes-responses


KevinOrfas (changes requested):
Tests pass, DoD reads as met. Three threads worth pausing on before merge.

**1. The single channel emitter, a 2.B call reaching its limit.**

Worth tracing the history because the answer is in the trace. PR #20 (Story 2.B) introduced the `ClientSocket` Nullable. Its `trackMessages()` emitted both `send()` and the server injection method on the same `'message'` event, by design. The review at the time named the cost:

> if you send a client message and then receive a server message, `tracker.data` will have both mixed together... worth thinking about whether the client should distinguish between outbound and inbound, otherwise the tracker won't be very useful beyond single action tests.

The call we made was to ship the single channel. Single action tests were the only consumer back then, so the directional question wasn't load bearing. Reasonable given what was in front of us.

3.C.1 is the first place that decision starts costing. `ConnectionManager` subscribes via the new `onMessage(listener)` channel, which sits on the same emitter. The manager now receives its own outbound subscribe messages on every test. `default: break` swallows them harmlessly today, but the moment any handler tightens or anyone asserts on the *full* set of received messages, it bites.

This is where we make the call we deferred in 2.B. Two channels, not one. `EVENT_OUTBOUND` and `EVENT_INBOUND`. `send` writes to outbound in both Real and Null. `onmessage` and `receiveMessage` write to inbound. `trackMessages` listens on outbound, `onMessage(listener)` listens on inbound. Symmetric across Real and Null. The ambiguity goes away.

When we hit an old shortcut like this, we close it in the PR that hits it, not as a follow up.

**2. Half the state in `connectionManager.ts` doesn't change behaviour.**

Three pieces of state appear to track which subscription cares about which collection:

- `collectionLiveSubscriptions: Map<collection, Set<subId>>`
- `collections: Set<string>` on each subscription state
- the loop in `markCollectionAsLive` that backfills both when a data message arrives

Both tests pass if you delete all three and replace `shouldRouteDataForCollection` with `return this.hasAnyLiveSubscription()`. The collection tied state has no effect on a routing decision either test makes.

It can't have an effect. Data messages on the wire carry `collection` but no sub id (`shared/protocol.ts:38` to `:55`). The client cannot tell which subscription a `{ type: 'added', collection: 'files', id: '1' }` belongs to, because the server hasn't told it. The bookkeeping is reaching for a discriminator that isn't on the wire.

The smallest shape that passes both tests:

```ts
private liveSubs = new Set<string>();

subscribe(name): SubscriptionHandle {
  const id = generateSubscriptionId();
  this.liveSubs.add(id);
  // ... rest same as before
}

stopSubscription(id) {
  this.liveSubs.delete(id);
  // ... rest same as before
}

private handleServerMessage(message) {
  if (message.type === 'added' || message.type === 'changed' || message.type === 'removed') {
    if (this.liveSubs.size === 0) return;
    this.store(message.collection).handleMessage(message);
    return;
  }
  if (message.type === 'ready') {
    this.subscriptions.get(message.id)?.readyResolver();
  }
}
```

One set, two state changes, one read at routing time. Same external behaviour as the current shape.

When per sub filtering becomes a real DoD ask later, the discriminator either lands on the wire (server tags data with sub id) or in the client (the client maps pub name to collection at subscribe time, then filters by collection on receive). Either choice is a different shape from what's there now, so landing the current bookkeeping means rewriting it whichever way that ticket goes.

**3. `handle.ready` hangs forever on a server `error`.**

PR #38 has the server emit `{ type: 'error', error: { code: 404, ... } }` when a client subscribes to a publication that wasn't `define()`d on the server. `connectionManager.ts:112` to `:134` doesn't handle `error`. The default branch swallows it. So `await handle.ready` for an unknown publication waits indefinitely.

There's a `readyRejector` captured at subscribe time, but the only path that calls it is the local `socket.send(...).catch(...)` block, i.e. failures to put bytes on the wire. Server side rejections never reach it.

The DoD says "subscribe sends message and routes responses." `error` is a response. A failing test:

```ts
it('rejects ready when subscribe fails on the server', async () => {
  const socket = ClientSocketWrapper.createNull();
  const server = socket.simulateServer();
  const manager = new ConnectionManager(socket);
  const handle = manager.subscribe('nope');
  const sent = socket.trackMessages().data[0] as SubscribeMsg;
  server.send({
    type: 'error',
    id: sent.id,
    error: { code: 404, message: 'Unknown publication' },
  });
  await expect(handle.ready).rejects.toMatchObject({ code: 404 });
});
```

Once #1 is sorted, `trackMessages().data[0]` will reliably be the subscribe rather than depending on send and receive interleaving.

KevinOrfas (changes requested):
**Top line**

The outbound/inbound split clears up the event model and the `StubbedClientSocket.onMessage → receiveMessage` rename resolves the name clash with the new wrapper method. Three new tests for the manager plus the rewrite of the inbound case in `clientSocket.test.ts` is solid test discipline.

I've pulled four threads I'd want us to walk through together before this lands, and codified most of them as tests against the PR branch so the behaviour is in plain view rather than me asserting it from the outside. There are also a handful of smaller things at the bottom.

---

**Thread 1: the routing test is passing by coincidence**

Have a look at the data branch in `connectionManager.ts`:

```ts
case 'added':
case 'changed':
case 'removed': {
  if (this.liveSubs.size === 0) return;
  this.store(message.collection).handleMessage(message);
  break;
}
```

It says "if no subscriptions are live, drop the data". The stop test in `tests/client/connectionManager.test.ts` runs with one subscription, so by the time the late `changed` arrives, `liveSubs.size` is 0 and the message is dropped. Green.

Drop this next to it:

```ts
it('late data for a stopped sub does not leak when another sub is still live', async () => {
  const socket = ClientSocketWrapper.createNull();
  const server = socket.simulateServer();
  const manager = new ConnectionManager(socket);
  const messages = socket.trackMessages();

  const filesHandle = manager.subscribe('files.all');
  const otherHandle = manager.subscribe('other.all');

  const filesSubId = (messages.data[0] as SubscribeMsg).id;
  const otherSubId = (messages.data[1] as SubscribeMsg).id;

  server.send({ type: 'added', collection: 'files', id: '1', fields: { name: 'existing.bam' } });
  server.send({ type: 'ready', id: filesSubId });
  server.send({ type: 'ready', id: otherSubId });
  await filesHandle.ready;
  await otherHandle.ready;

  filesHandle.stop();

  server.send({ type: 'changed', collection: 'files', id: '1', fields: { name: 'late.bam' } });

  expect(manager.store('files').getById('1')).toEqual({ _id: '1', name: 'existing.bam' });
});
```

Red. `otherHandle` is still live so `liveSubs.size > 0`, the late `changed` flows through, and the doc becomes `late.bam`.

What is the guard actually defending? `DataMsg` in `shared/protocol.ts:38-55` carries `collection` and `id` but no sub id. The manager genuinely cannot route data per subscription from its own state. So the answer probably lives somewhere else:

- The server stops sending data for a sub id once it sees `unsubscribe`. The client guard then has nothing to do. This is what real DDP does and feels like publications territory rather than manager territory.
- Or the protocol grows a sub id on `DataMsg` and the manager filters. Bigger change.

Which way were you heading? If it's the server side fence, I'd drop the `liveSubs.size === 0` check from here and let the failing test above sit as a contract test for whichever ticket picks the fence up. The guard as written reads like a promise the manager can't actually keep.

---

**Thread 2: `stopSubscription` leaks the subscription entry**

Walking the state in `connectionManager.ts`:

|                          | `subscriptions` | `liveSubs` |
|--------------------------|-----------------|------------|
| `subscribe()`            | add             | add        |
| `stopSubscription()`     | untouched       | delete     |
| `error` message          | delete          | delete     |
| subscribe send rejects   | delete          | delete     |

After `handle.stop()`, the entry stays in `subscriptions` with its resolver and rejector still wired up. If a slow `ready` or `error` arrives later for that id, we call those on an already settled promise. JavaScript no ops it, fine, but the entry is now dead weight that never goes away. Long lived page plus lots of subs, that grows.

A small inspection seam plus this test pins it down:

```ts
it('stop releases subscription bookkeeping', () => {
  const socket = ClientSocketWrapper.createNull();
  const manager = new ConnectionManager(socket);
  const handle = manager.subscribe('files.all');
  handle.stop();
  expect(manager.activeSubscriptionCount()).toBe(0);
});
```

There's a related observation underneath this: `liveSubs` is always a subset of `subscriptions.keys()`, so the two maps are tracking the same thing twice and that's what created room for them to diverge. Collapsing to one map (or a status field on `SubscriptionState`) removes the divergence at the source. The `id` field inside `SubscriptionState` is also unused, the map key already carries it.

What shape do you fancy here, single map with status, or just delete on stop and let absence mean gone?

---

**Thread 3: `isServerMessage` accepts anything with a `type` key**

`clientSocket.ts:4-6`:

```ts
function isServerMessage(data: unknown): data is ServerMessage {
  return typeof data === 'object' && data !== null && 'type' in data;
}
```

`ServerMessage` is a union of six literals (`'ready' | 'added' | 'changed' | 'removed' | 'result' | 'error'`, `shared/protocol.ts:79`). The guard claims to narrow to that union but only checks that *some* `type` exists. So `{ type: 'totally-bogus' }` slips through, the cast `listener(data as ServerMessage)` in `RealClientSocket.onmessage` carries it into the manager, and the `default: break` in `handleServerMessage` swallows it without a peep.

Quick jargon: a *type guard* is a function whose signature ends `data is X`. The compiler trusts the signature. The body has to prove what the signature claims, or the rest of the system is reading the type on faith.

If we export the helper, this test catches it cleanly:

```ts
expect(isServerMessage({ type: 'totally-bogus' })).toBe(false);
expect(isServerMessage({ type: 'ready', id: '1' })).toBe(true);
```

Two shapes worth considering: narrow on the literal set (cheapest), or per branch schema checks that also verify `collection`/`id`/`fields` for data messages (stronger frontier with the server). Which one fits the level of trust we want here?

---

**Thread 4: `ConnectionManager` never unhooks from the socket**

In the constructor:

```ts
this.socket.onMessage((message) => {
  this.handleServerMessage(message);
});
```

`onMessage` returns a teardown function. Look at `clientSocket.ts:108-115`:

```ts
onMessage(listener: (message: ServerMessage) => void): () => void {
  const handler = (data: unknown) => listener(data as ServerMessage);
  this.emitter.on(EVENT_INBOUND, handler);
  return () => { this.emitter.off(EVENT_INBOUND, handler); };
}
```

That return value is dropped on the floor here, so the manager pins itself (and every `ReactiveStore` it has handed out) in memory for as long as the socket lives. The pattern we want is already in `ReactiveStore.onChange` at `reactiveStore.ts:84-90`: caller keeps the teardown, runs it when done. Same shape works here, surfaced as `manager.dispose()` or `manager.close()`.

Failing test that drives both the API and the wiring:

```ts
it('after dispose, server data does not mutate manager stores', async () => {
  const socket = ClientSocketWrapper.createNull();
  const server = socket.simulateServer();
  const manager = new ConnectionManager(socket);
  const messages = socket.trackMessages();

  const handle = manager.subscribe('files.all');
  server.send({ type: 'ready', id: (messages.data[0] as SubscribeMsg).id });
  await handle.ready;

  manager.dispose();

  server.send({ type: 'added', collection: 'files', id: 'late', fields: { name: 'x.bam' } });

  expect(manager.store('files').getById('late')).toBeUndefined();
});
```

Red today, no `dispose`. Same question as thread 2 really, does the manager own a lifecycle?

---

**Smaller threads, still in this PR**

- The event split made `trackMessages()` outbound only, but nothing pins it down. A test that `tracker.data` stays empty after `server.send(...)` locks it. `tests/client/clientSocket.test.ts` is the right home.
- `generateSubscriptionId` has a counter fallback if `crypto.randomUUID` is missing. What environment are we shipping to that lacks it? Modern browsers and Node 19+ ship it. If the honest answer is "none we care about", the fallback is untested code that quietly produces colliding `sub-1` ids across managers in the same module realm. Either drop it or test it. Which?
- `stopSubscription` logs send failures with `console.error` while `subscribe` rejects `ready` on failure. Is `stop()` meant to be fire and forget, or should it return a promise the caller can await? Both are defensible, the asymmetry just feels worth making deliberate. Which were you reaching for?
- `default: break` in `handleServerMessage` quietly drops `'result'` messages (method results, future ticket). When method work lands we'll forget this case unless either (a) the switch narrows so `result` doesn't typecheck as fallthrough, or (b) we leave an `assertNever` branch. `ReactiveStore.handleMessage` does option (b) at `reactiveStore.ts:65-67` if you want a mirror.

---

**Non blocking**

- Test step comments like `// Manager sent the subscribe message.` echo what `it(...)` already says. Trim if you like.

KevinOrfas (changes requested):
Solid work, and it lands the subscribe flow cleanly. Before the things to change, the part worth calling out: the inbound/outbound split is the standout. Before this PR, `trackMessages()` and `onMessage()` both sat on a single `'message'` event, so a test asking "what did the client send" would also catch whatever the server sent. Splitting them into `'outbound'` and `'inbound'` separates the two questions, and the new `outPutTracker tracks messages sent by the client only` test pins it so it stays separated. The red/green cadence reads clearly in the commits, and `isServerMessage` as a guard at the socket edge is the right instinct.

Two things to walk through here, both needed before merge, each with a red test that drives it. The third strand, the manager's lifecycle and `dispose()`, we're settling on the Thread 4 discussion separately, so it isn't repeated below.

### 1. `handleServerMessage` throws on `result` — `connectionManager.ts:139`

Walk into the switch. Five of the six branches are gentle when they can't act: `ready` and `error` for a sub id the client doesn't have just fall through and return. `result` does something different, it throws.

Follow the throw. `handleServerMessage` doesn't run on its own stack; it runs inside the emitter's dispatch loop. Look at `EventEmitter.emit` in `server/infrastructure/outputTracker.ts:15`, `for (const handler of [...handlers]) { handler(data); }`. A throw out of one handler breaks that `for`, so any other listener on `'inbound'` stops receiving that message. In the real socket the same throw lands in `socket.onmessage` with nothing above it to catch it.

And it does fire for real: `isServerMessage` (`clientSocket.ts:20`) waves `result` messages through as valid, so a `result` from the server reaches this line.

The answer already lives two cases up in the same switch. The codebase has a vocabulary for "received it, can't act on it, that's fine": `ready` for an unknown id does exactly that, and `reactiveStore.ts` has the richer version with its `'skipped'` / `'failed'` observer outcomes. `result` belongs in that family. `case 'result': break;` is enough.

Keep it as a named case though, don't delete it. The `default: assertNever(message)` line is an exhaustiveness check: TypeScript narrows the union as each case is handled, and if every member is covered, `message` in `default` has type `never` and compiles. Miss one and it won't be `never`, and the build fails. A named `case 'result': break;` keeps that guarantee; deleting the case breaks it.

**Red test** (`connectionManager.test.ts`): simulate the server sending a `result` message, assert `server.send(...)` does not throw, then send an `added` and assert it still routes into the store. The second half is the real test. Anyone can stop a throw; what you need to prove is that the inbound dispatch loop survived the message and the next one still lands.

### 2. `handle.stop()` tells the server to stop, but the client's view never clears

`handle.stop()` sends `unsubscribe` and drops the subscription from the manager's map. Ask the lifecycle question: what happens to the documents that subscription put in the store?

Walk the client first. `ConnectionManager` receives `added` / `changed` / `removed` and routes them by `collection` into a `ReactiveStore`. There's no per-subscription document tracking, and there can't be: `DataMsg` carries a `collection` and a document `id`, never a subscription id. So the client cannot attribute a document to the sub that brought it in, which means `stop()` cannot purge the store itself, and shouldn't try. A document covered by two subscriptions would vanish the moment either one stopped.

So the client does the only correct thing: it sends `unsubscribe` and waits for the server to tell it what to drop. The client code in this PR is right.

Now walk the server. `server/logic/publications.ts`, the `unsubscribe` branch at line 114. It calls `cleanup()` to stop the watcher and deletes the map entry. It sends the client nothing. No `removed`. So the client waits for a message that never comes, and the documents sit in the store forever. The subscribe flow this PR builds isn't finished until the server closes the loop, so the fix lands here, in this PR.

The pieces are already around that `unsubscribe` branch:

- The `subscribe` path already has the documents in hand: `docs` from `mongo.find`, plus every `insert` the watcher pushes afterwards.
- The per-client, per-sub map is already the place to hang state. Today its value is a bare `cleanup` function. It needs to also remember the document ids that subscription has sent `added` for.
- `publications.ts` already has `addedMessage` and `readyMessage` helpers at the top of the file. This wants a `removedMessage` sibling.

On `unsubscribe`: call `cleanup` as it does now, then send a `removed` for each id that subscription sent. The client already routes `removed` through `ReactiveStore.handleMessage`, so the client side needs no new code, it just starts receiving the messages it was already waiting for.

One decision to make while you're in here, not after. If a document is covered by two subscriptions, sending `removed` when the first one stops is wrong, the second sub still wants it. There is one publication per collection today, so the clash is latent rather than live, but it is a real correctness call. Either refcount the document ids across a client's subscriptions and only send `removed` when the last sub covering an id stops, or accept the one-publication-per-collection limit knowingly with a comment that says so. `learning/unsubscribe-is-the-rare-path.md` walks the same "who owns subscription state" fault line and is worth a read before you pick.

**Red test**: one, and it's server-side, because the fix is server-side. A `publications` test that subscribes, unsubscribes, and asserts the client receives `removed` for the documents it was sent. That's red now and goes green when the server fix lands; it's the test that drives the change.

The client side needs no code and no red test. `ConnectionManager` already routes `removed` into the store, covered by the 3.B.3 routing tests. Once the server half is in, it's worth adding one `connectionManager` test that runs the whole loop, subscribe, stop, server sends `removed`, store cleared, as a regression guard. It'll be green the first time it runs, and that's the point of it: it locks the two halves together so a later change can't quietly break the loop. Add it as a guard, not as a red-first step.

### Smaller things

- In `isServerMessage`, the `removed` case body is identical to `added` / `changed` (`clientSocket.ts:15-19`); fold it into the same fallthrough group.
- `assertNever` is defined identically in `connectionManager.ts:15` and `reactiveStore.ts:12`, a candidate for `shared/`.
- `StubbedServer.messages` (`clientSocket.ts:179`) is declared but never written or read. Already present before this PR, but this PR touches `StubbedServer`, so cheap to clear out while you're in there.

### Question for the pair

Both gaps sit on the same line. The subscribe happy path is solid; each gap is on a rarer path: a message type not handled yet, the stop half of subscribe/stop. The lifecycle strand on Thread 4 is a third one on the same line, and it threw up something worth carrying into this question: a rare path can have a test that's green for a reason other than the one the test name claims.

So sketch it: the full message lifecycle of one subscription, birth to teardown, every arrow between client and server. Which arrows have a test, which are just assumed to work, and which have a test that passes for a reason you didn't intend? The rare paths are where this PR's gaps clustered, and they will keep clustering there until each arrow is pinned for the right reason.

KevinOrfas (changes requested):
**Top line**

The `result` case as a named break, the error rejection test, the server side `removed` loop on unsubscribe, the outbound/inbound split, the per branch validation in `isServerMessage`: all of these close what the prior reviews opened. Worth saying that's solid work and reads cleanly.

The rest of this review hangs on one move. The PR closes the happy path of subscribe. Each thread below is a rare path that the happy path tests don't touch. The shape is the same on each one: small failing test, you run it on the branch, you reply with what you saw in your own words. Once we've named what surfaces, the fix lands in this PR.

---

### Thread 1: the guard at the edge

Open `tests/client/clientSocket.test.ts`. Add these three next to the existing `isServerMessage` cases:

```ts
it('rejects undefined', () => {
  expect(isServerMessage(undefined)).toBe(false);
});

it('rejects an array shaped like a message', () => {
  const arrayShapedAsMessage: unknown = Object.assign([], { type: 'ready', id: '1' });
  expect(isServerMessage(arrayShapedAsMessage)).toBe(false);
});

it('rejects a removed message carrying unexpected extras', () => {
  expect(
    isServerMessage({ type: 'removed', collection: 'files', id: '1', fields: {} }),
  ).toBe(false);
});
```

Run `npx vitest run tests/client/clientSocket.test.ts`. Reply with which went green, which went red, and which one surprised you.

Two related things to walk into once you've reported back, both in the same function.

The cast `data as Record<string, unknown>` on line 7 is what one of those red points at. A cast tells the compiler \"trust me, this is shaped like X\" and the runtime doesn't verify. So whatever leaks through the first guard gets full string-keyed access on the way to the per-case checks. Worth deciding whether the cast belongs at all, or whether each property read narrows with `in` and drops the wholesale assertion.

The fall through is the other thing:

```ts
case 'ready':
case 'result':
  return typeof msg.id === 'string';
```

The two cases share a check today, but mean different things. `ready` says a subscription has started. `result` says a method call has finished. If `result` ever needs its own check (a `result` field on the message, for instance), the first thing the next change has to do is split the fold. Cheaper to give each case its own line now.

Separately, once the per-case checks grow, the whole switch starts wanting to be a small table of validators keyed on `type`. Worth pinning that for another piece rather than landing it here.

One thing to lock in a comment over \`isServerMessage\` once you've decided: is the edge strict (reject extras) or permissive (read what we need)? Strict catches schema drift early. Permissive lets the server move ahead of the client. The current behaviour is permissive but isn't written down anywhere.

---

### Thread 2: the manager that resurrects

Add `tests/client/connectionManager.dispose.test.ts`:

```ts
import { describe, expect, it } from 'vitest';
import { ClientSocketWrapper } from '../../client/clientSocket.ts';
import { ConnectionManager } from '../../client/connectionManager.ts';

describe('ConnectionManager — after dispose', () => {
  it('store() throws', () => {
    const socket = ClientSocketWrapper.createNull();
    const manager = new ConnectionManager(socket);
    manager.dispose();
    expect(() => manager.store('files')).toThrow();
  });

  it('subscribe() throws', () => {
    const socket = ClientSocketWrapper.createNull();
    const manager = new ConnectionManager(socket);
    manager.dispose();
    expect(() => manager.subscribe('files.all')).toThrow();
  });
});
```

Run it. Reply with what you saw.

When you walk into why, the `disposed` flag is already in the file. The inbound listener reads it in the constructor closure. The two callers that resurrect state, `store()` and `subscribe()`, don't. One private `assertNotDisposed()` is called at the top of both closes it. Drop a sentence in a code comment naming why the flag is consulted in three places once the fix lands, so the next person reading sees the shape and doesn't have to derive it.

---

### Thread 3: the socket the manager doesn't watch

Add `tests/client/connectionManager.socketClose.test.ts`:

```ts
import { describe, expect, it } from 'vitest';
import { ClientSocketWrapper } from '../../client/clientSocket.ts';
import { ConnectionManager } from '../../client/connectionManager.ts';
import { SubscribeMsg } from '../../shared/protocol.ts';

it('disposes when the socket closes, without an explicit dispose call', async () => {
  const socket = ClientSocketWrapper.createNull();
  const server = socket.simulateServer();
  const manager = new ConnectionManager(socket);
  const messages = socket.trackMessages();

  const handle = manager.subscribe('files.all');
  server.send({ type: 'ready', id: (messages.data[0] as SubscribeMsg).id });
  await handle.ready;

  await socket.close();

  expect(manager.activeSubscriptionCount()).toBe(0);
});
```

Run it. Reply with what you saw.

The server already has this shape. In the `Publications` constructor at `server/logic/publications.ts:35`, it subscribes to `this.ws.onDisconnect`, and when the connection drops, the client's subscriptions get torn down off the back of that event. The connection owns the lifecycle, the cleanup hangs off the disconnect. The client wants the same arrangement.

For that to land, two things have to change.

First, the socket wrapper needs to surface a close event, because today nothing on it tells the manager that the socket has gone. Add a method to `ClientSocketWrapper`:

```ts
onClose(listener: () => void): () => void
```

In `RealClientSocket`, wire that listener to the underlying `socket.onclose`. In `StubbedClientSocket`, expose a `simulateClose()` on the stubbed server so a test can fire a close on demand, the same way `simulateServer().send(...)` fires inbound messages today.

Second, the manager subscribes to it. In the constructor, alongside the inbound listener it already sets up, add:

```ts
this.teardownCloseListener = this.socket.onClose(() => this.dispose());
```

`dispose()` already guards itself with the `disposed` flag, so if a caller later calls `dispose()` explicitly after the socket has already closed, the second call is a no op. Same path, two ways in.

That last point is where threads 2 and 3 turn out to be the same conversation. Once the close listener is wired and the `disposed` flag is also consulted by `store()` and `subscribe()` (thread 2), the manager has one cleanup path (`dispose()`), two ways to reach it (the caller, or the socket closing), and three places where `disposed` is read (the inbound listener, `store()`, `subscribe()`). Worth drawing that on paper in your reply before the code goes in, so the whole shape lands in one move rather than two patched in series.

---

### Thread 4: the shared document

Add `tests/logic/publications.refcount.test.ts`:

```ts
import { describe, expect, it } from 'vitest';
import { Publications } from '../../server/logic/publications.ts';
import { MongoWrapper } from '../../server/infrastructure/mongo.ts';
import { WebSocketWrapper } from '../../server/infrastructure/websocket.ts';

it('unsubscribing one sub does not remove a doc the other still covers', async () => {
  const mongo = MongoWrapper.createNull({
    find: [
      [{ _id: '1', name: 'shared.bam' }],
      [{ _id: '1', name: 'shared.bam' }],
    ],
  });
  const pubs = new Publications(mongo, ws);

  pubs.define('files.all', () => ({ collection: 'files', query: {} }));
  pubs.define('files.featured', () => ({ collection: 'files', query: {} }));

  await pubs.handleMessage(client.id, { type: 'subscribe', id: 'subA', name: 'files.all' });
  await pubs.handleMessage(client.id, { type: 'subscribe', id: 'subB', name: 'files.featured' });

  const countBeforeUnsub = client.messages.length;
  await pubs.handleMessage(client.id, { type: 'unsubscribe', id: 'subA' });

  const removedForSharedDoc = client.messages
    .slice(countBeforeUnsub)
    .find((m) => (m as { type: string }).type === 'removed');

  expect(removedForSharedDoc).toBeUndefined();
});
```

Run it. Reply with what you saw.

Then walk into the call. This is the decision you and I flagged together. Two paths. One, accept the limit of one publication per collection knowingly and lock it in a short comment at the top of the unsubscribe branch so the next change reads it. Two, refcount document ids per client across that client's subscriptions and only send `removed` when the last sub covering an id stops. Stricter, bigger change.

My read for this PR: ship the comment, defer the refcount until a second publication over a single collection is on the roadmap. The red test stays in the tree as the guard. Either it goes green when refcount lands, or it stands as the contract saying \"the day this goes green, the assumption in that comment broke.\" Happy to take the other call if you'd rather close it now.

---

### Thread 5: the home of `assertNever`

Not a test, a file path.

`client/connectionManager.ts:4` and `client/reactiveStore.ts:3` import `assertNever` from `../tests/shared/helperFunctions.ts`. The duplication the earlier review flagged is gone, which is the good half. The shared helper landing in the test tree is the half to revisit. Production code reaching into tests inverts the dependency arrow. The day a tsconfig split excludes `tests/**` from the client build, or a bundler prunes the test tree, this breaks.

Right home is `shared/`, sibling to `protocol.ts`. Two file moves, three import path swaps.

---

### What the four reds share

Each thread above is a rare path that the happy path tests don't touch. An unexpected message shape. A teardown trigger. An overlap case. The fix on each is small once the test is in. The pattern worth taking from this PR into the next ticket: when a rare path shows up in DoD, the failing test for it lands before the production code, and the rare path comes in on the same rails as the happy path.

Run the four. Drop a reply with what each one shows in your own words. Once we've named what surfaces, the fixes land in this PR.

KevinOrfas (changes requested):
Thanks for the work on this round, Ebuka. A lot has landed well. The dispose lifecycle reads cleanly with the disposed flag wired into the three places we talked about, moving assertNever out of the test folder and into shared feels right, splitting ready and result in the switch is exactly the shape I was hoping for, the handle.ready rejection when stop runs before the server responds matches the symmetry we wanted, and the removed messages on unsubscribe close the loop we left open last review. Tests are green and the one expected fail is doing what we want it to do. Two things I'd like us to fold into this PR before merge so we don't carry them as follow ups, plus a smaller clarification and the usual end of review odds and ends.

**On the socket close path.** The stub side reads clearly. simulateClose emits the disconnection event, the listener fires, dispose runs, all sensible. When I traced the same flow through RealClientSocket though I think the production half tells a different story. In connect I can see onopen, onerror and onmessage being assigned but socket.onclose only gets set inside close itself, on line 152, which only runs when we initiate the close. The WebSocket default for onclose is null, so when the server sends a close frame the event is absorbed and the manager never auto disposes.

To make sure I was reading this right I wrote a small test against a real ws server. It currently fails on this branch with activeSubscriptionCount staying at 1 after the server side close. Here it is, drop it in at `tests/client/connectionManager.realSocketClose.test.ts`:

```ts
import { describe, expect, it } from 'vitest';
import { WebSocketServer } from 'ws';
import { ClientSocketWrapper } from '../../client/clientSocket.ts';
import { ConnectionManager } from '../../client/connectionManager.ts';

describe('ConnectionManager with a real socket', () => {
  it('disposes when the server closes the socket', async () => {
    const server = new WebSocketServer({ port: 0 });
    const port = (server.address() as { port: number }).port;

    const socket = ClientSocketWrapper.create(`ws://localhost:${port}`);
    await socket.connect();
    const manager = new ConnectionManager(socket);

    manager.subscribe('files.all');

    await new Promise<void>((resolve) => setTimeout(resolve, 50));

    for (const client of server.clients) {
      client.close();
    }

    await new Promise<void>((resolve) => setTimeout(resolve, 100));

    expect(manager.activeSubscriptionCount()).toBe(0);

    server.close();
  });
});
```

The fix on the production side is small. Move the onclose wiring into connect so it gets set the moment the socket exists, then have close just wait for the disconnection event rather than installing its own handler. Roughly this shape, tweak as you see fit:

```ts
// inside connect, alongside onopen/onerror/onmessage
this.socket.onclose = () => {
  this.emitter.emit(CLIENT_DISCONNECTION_EVENT);
};

// close becomes
close(): Promise<void> {
  return new Promise((resolve) => {
    const socket = this.socket;
    if (!socket || socket.readyState === WebSocket.CLOSED) {
      resolve();
      return;
    }
    const teardown = this.onClose(() => {
      teardown();
      resolve();
    });
    socket.close();
  });
}
```

One small bookkeeping thing while you're in there. The test imports ws directly because the test needs a WebSocket server and Node only ships the client side of the spec as a global. ws is already in node_modules through @fastify/websocket but isn't declared as a direct devDependency, so we're piggybacking on a transitive resolution that could shift if Fastify ever swaps the underlying library. Worth adding ws to devDependencies in this same commit so the test owns the dependency it actually uses.

**On the cast inside onMessage.** Both RealClientSocket and StubbedClientSocket build the listener with `const handler = (data: unknown) => listener(data as ServerMessage)`. The runtime contract is honoured because isServerMessage runs upstream of the emit, but the type system can't see that and we end up with the same cast in two places. I'd rather close this in the same PR than leave it for later. The leanest move I can see is to lift the handler creation into a tiny module local helper that narrows with the existing guard, which both kills the cast and dissolves the duplication:

```ts
function listenForServerMessage(
  emitter: EventEmitter,
  listener: (message: ServerMessage) => void,
): () => void {
  const handler = (data: unknown) => {
    if (isServerMessage(data)) {
      listener(data);
    }
  };
  emitter.on(EVENT_INBOUND, handler);
  return () => emitter.off(EVENT_INBOUND, handler);
}
```

Then both onMessage implementations collapse to a single line that calls this helper. No cast, single source of truth for the inbound listening shape. There's a redundant isServerMessage call inside that handler given the upstream filter, but it's a cheap shape check and it's defensive in the stub path where receiveMessage takes a typed input but the emitter is still erased to unknown by the time the listener sees it. Open to a different shape if you've got one in mind that you like better.

**On isServerMessage.** The comment at the top of clientSocket.ts says the switch should be permissive, read what we need, and most cases line up with that. The removed case caught my eye because it carries that extra check, no fields in msg, which actively rejects extras. I can see why that might be deliberate, a removed carrying fields is the server contradicting itself and probably worth surfacing rather than letting through silently. I can also see a world where the check just landed there and the comment is the truth. Which is closer to what you had in mind? Either works for me, I'd just like the code and the comment to be telling the same story so the next reader doesn't have to guess.

**On the refcount limitation.** The comment at the top of unsubscribe and the it.fails test together feel like the right shape for deferring this. CI still runs the case and would shout if someone made it pass without dropping the marker. Wanted to flag I'm reading our earlier conversation as us agreeing this stays as a documented limit for now, with the proper refcount work scoped when a second publication arrives. Shout if you saw it differently.

**One small thing.** The constant CLIENT_DISCONNECTION_EVENT on line 48 is spelled disconnetion, missing the c. Easy to copy forward so figured I'd flag it while we're here.

Thanks again, this has moved a lot.

KevinOrfas (approved):
This reads exactly the way I was hoping. Onclose moved into connect, close refactored cleanly around the disconnection event, the listenForServerMessage helper kills the cast and collapses the duplication in one go, ws is declared properly in devDependencies, the comment on the guard now matches what the code actually does, and the typo is gone. All green locally including the new real socket test.

One thing worth calling out positively. The handle.ready rejection check you added inside the new test (the `await expect(handle.ready).rejects.toThrow(...)` line) is a better test than the one I sketched. It pins down not just that the active count drops but that the ready promise chain actually fires the rejection we built into stopSubscription. Nice catch.

One small thing you can take or leave, not blocking. The new test file is named realsocketclose rather than realSocketClose, which sits a little oddly next to the camelCase pattern used by socketClose.test.ts and the others. Easy to rename on a future commit if you want, no rush.

Happy to merge. Thanks for closing this one out properly.

PR #46, opened 2026-05-22
Eko 272/3c2 subscribe with params server side filtering by client knowledge

Iam-Ebuka:
- test: red - passes subscribe params to the publication query
- added the solution in publications for the test to pass
- test: green - subscribe with params includes them in the outbound message
- docs: commented why we should leave params type as it is

https://linear.app/ekohacks/issue/EKO-272/3c2-subscribe-with-params-server-side-filtering-by-client-knowledge

### DIFF
--- a/server/logic/publications.ts
+++ b/server/logic/publications.ts
@@ -1,4 +1,5 @@
 import { ClientMessage, PublicationsObserver, PublicationsReasons } from '../../shared/protocol.ts';
+import { hasMongoOperator } from '../../shared/helperFunctions.ts';
 import { MongoWrapper } from '../infrastructure/mongo.ts';
 import { WebSocketWrapper } from '../infrastructure/websocket.ts';
 import { ChangeEvent } from '../../shared/types.ts';
@@ -67,6 +68,20 @@ export class Publications {
     }
   }
 
+  private sendPublicationError(
+    clientId: string,
+    message: ClientMessage,
+    reason: PublicationsReasons,
+    messageText: string,
+  ): void {
+    this.ws.send(clientId, {
+      type: 'error',
+      id: message.id,
+      error: { code: 400, message: messageText },
+    });
+    this.notifyObserver(message, 'failed', reason);
+  }
+
   private tearDownClient(clientId: string): void {
     const clientSubs = this.subscriptions.get(clientId);
     if (!clientSubs) return;
@@ -96,30 +111,14 @@ export class Publications {
         return Promise.resolve();
       }
 
-      const hasMongoOperator = (obj: unknown): boolean => {
-        if (obj === null || obj === undefined) return false;
-        if (Array.isArray(obj)) return obj.some((item) => hasMongoOperator(item));
-        if (typeof obj === 'object') {
-          for (const [key, val] of Object.entries(obj as Record<string, unknown>)) {
-            if (key.startsWith('$')) return true;
-            if (hasMongoOperator(val)) return true;
-          }
-        }
-        return false;
-      };
-
       if (message.params && hasMongoOperator(message.params)) {
-        this.ws.send(clientId, {
-          type: 'error',
-          id: message.id,
-          error: {
-            code: 400,
-            message: `Invalid subscription params: mongo operators are not allowed`,
-          },
-        });
-
-        this.notifyObserver(message, 'failed', 'invalid-params');
-        return Promise.resolve();
+        this.sendPublicationError(
+          clientId,
+          message,
+          'invalid-params',
+          'Invalid subscription params: mongo operators are not allowed',
+        );
+        return;
       }
 
       let collection: string;
@@ -133,14 +132,8 @@ export class Publications {
             ? `Publication query failed: ${err.message}`
             : 'Publication query failed';
 
-        this.ws.send(clientId, {
-          type: 'error',
-          id: message.id,
-          error: { code: 400, message: messageText },
-        });
-
-        this.notifyObserver(message, 'failed', 'publication-query-failed');
-        return Promise.resolve();
+        this.sendPublicationError(clientId, message, 'publication-query-failed', messageText);
+        return;
       }
 
       let docs;
@@ -152,14 +145,8 @@ export class Publications {
             ? `Publication query failed: ${err.message}`
             : 'Publication query failed';
 
-        this.ws.send(clientId, {
-          type: 'error',
-          id: message.id,
-          error: { code: 400, message: messageText },
-        });
-
-        this.notifyObserver(message, 'failed', 'publication-query-failed');
-        return Promise.resolve();
+        this.sendPublicationError(clientId, message, 'publication-query-failed', messageText);
+        return;
       }
 
       const documentIds = new Set<string>();
--- a/server/logic/publications.ts
+++ b/server/logic/publications.ts
@@ -73,11 +73,12 @@ export class Publications {
     message: ClientMessage,
     reason: PublicationsReasons,
     messageText: string,
+    code: number,
   ): void {
     this.ws.send(clientId, {
       type: 'error',
       id: message.id,
-      error: { code: 400, message: messageText },
+      error: { code: code, message: messageText },
     });
     this.notifyObserver(message, 'failed', reason);
   }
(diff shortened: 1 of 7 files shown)


KevinOrfas (commented):
Ebuka, two red tests for the params work. Both stop at red. The green is yours if you agree the red is right.

Starting from the comment in `publications.ts` about why `params` stays as `Record<string, unknown>`. That comment makes the boundary explicit, and the boundary is what the first red is asking about.

```ts
it('rejects subscribe params that smuggle mongo operators', async () => {
  const mongo = MongoWrapper.createNull({ find: [[]] });
  const pubs = new Publications(mongo, ws);

  let queryFnCalled = false;
  pubs.define('files.byFolder', (params) => {
    queryFnCalled = true;
    const folderId = (params as { folderId: string }).folderId;
    return { collection: 'files', query: { folderId } };
  });

  await pubs.handleMessage(client.id, {
    type: 'subscribe',
    id: 'sub1',
    name: 'files.byFolder',
    params: { folderId: { $ne: null } },
  });

  expect(queryFnCalled).toBe(false);
  expect(client.messages).toContainEqual(
    expect.objectContaining({ type: 'error', id: 'sub1' }),
  );
});
```

Sends `params: { folderId: { $ne: null } }` to a publication whose body is the obvious `{ folderId: params.folderId }`. Asserts `queryFn` was never called, and the client received an `error`. Today both fail because nothing on the path between the wire and `queryFn` intercepts the operator shape, and Mongo would happily return every doc with a folderId.

The test is loose about *where* the rejection sits. What I'd like is your reading of the right shape before you write the green.

```ts
it('sends an error to the client when the publication query throws', async () => {
  const mongo = MongoWrapper.createNull({ find: [[]] });
  const notifications: { outcome: ObserverOutcome; reason?: string }[] = [];
  const observer: PublicationsObserver = {
    onMessage(_msg, outcome, reason) {
      notifications.push({ outcome, reason });
    },
  };
  const pubs = new Publications(mongo, ws, observer);

  pubs.define('files.byFolder', (params) => {
    if (typeof (params as { folderId?: unknown }).folderId !== 'string') {
      throw new Error('folderId is required and must be a string');
    }
    return { collection: 'files', query: {} };
  });

  await pubs.handleMessage(client.id, {
    type: 'subscribe',
    id: 'sub1',
    name: 'files.byFolder',
    params: {},
  });

  expect(client.messages).toContainEqual(
    expect.objectContaining({ type: 'error', id: 'sub1' }),
  );
  expect(notifications).toContainEqual(expect.objectContaining({ outcome: 'failed' }));
});
```

A publication that defensively throws on bad params. The throw escapes `handleMessage` as a rejected promise, so the client sees nothing and the observer isn't notified. Compare with the unknown-publication branch right above in the source, which sends `error` and notifies `failed` cleanly.

This test is more opinionated than the first. It asserts the server catches and translates, symmetric to that neighbour. If you'd rather throws bubble for some reason, the test is wrong and we change it. Otherwise the green is whatever shape feels right to you, including whether "publication threw" earns its own `PublicationsFailReason`.

Two smaller things. `docs/websocket-architecture/er-diagram.md:81` still says params are method only and array shaped. Is the diagram living, or do we revisit later? And subscribing twice with the same id and new params is now the implicit "change params" path. Duplicate sub id branch handles teardown so it works, just flagging it isn't named anywhere.

server/logic/publications.ts
```diff
@@ -94,8 +96,60 @@ export class Publications {
         return Promise.resolve();
       }
 
-      const { collection, query } = queryFn();
-      const docs = await this.mongo.find<{ _id: string }>(collection, query);
+      if (message.params && typeof (message.params as { folderId: string }).folderId !== 'string') {
```
KevinOrfas: Ebuka, this hangs together and the red/green steps are clean to follow. I want to think through one thing with you, starting from your own tests, because I reckon they already contain the answer between them.

Take the second test, the one where the publication throws:

```ts
pubs.define('files.byFolder', (params) => {
  if (typeof (params as { folderId?: unknown }).folderId !== 'string') {
    throw new Error('folderId is required and must be a string');
  }
  return { collection: 'files', query: {} };
});
```

I like this a lot. The publication decides what `folderId` means and complains when it's wrong. That's the publication's job, it owns its own shape. The engine just has to carry the complaint back to the client, which your `try/catch` does.

Now read it next to what the engine does a few lines up, right here:

```ts
if (message.params && typeof (message.params as { folderId: string }).folderId !== 'string') {
  // "Publication query failed: folderId is required and must be a string"
}
```

Here the engine itself knows there's a thing called `folderId`. Two places now hold the same rule, and the engine is the generic one, it sits under every publication anyone will ever write.

A question to sit with: what happens when I define `files.byUser` keyed on `userId` and subscribe with `{ userId: 'alice' }`? I tried it. The engine turns it away with "folderId is required", before my publication gets a look in. So the engine isn't validating params in general, it's enforcing one publication's field on all of them.

And here's the bit I'd really like your eye on. Trace the throwing test by hand. It subscribes with `params: {}`. Which branch catches it first, the engine's `folderId` check, or your publication's `throw`? Walk it and watch where it returns. Then the same trick on the operator test: it sends `{ folderId: { $ne: null } }` and goes green, but ask it the harder version, `{ folderId: 'folder-a', status: { $ne: null } }`, an operator hiding in a key that isn't `folderId`. Does the `$ne` still reach Mongo? That second case is closer to what "server side filtering by client knowledge" is actually protecting against, an attacker doesn't politely put their operator where we're looking.

So the shape I'd nudge towards: what if the publication owns "is `folderId` a string", since yours already does, and the engine takes the one job that's genuinely generic, refusing any param key that starts with `$`, at any depth, before `queryFn` runs? Then the engine never needs to know the word `folderId`. But that's me reading the boundary one way. Where do you think it sits?

And the smaller thread, live not later: when params get turned away you call `notifyObserver(..., 'publication-threw')`, but nothing threw. What would you call that event, if it's not a throw?

KevinOrfas: This is exactly it. The gate reads any `$`-prefixed key at any depth now, so the smuggle I was worried about, `{ folderId: 'a', status: { $ne: null } }`, gets turned away, and a publication keyed on anything other than folderId just works. And splitting the reason into `invalid-params` versus `publication-query-failed` answers the other thread cleanly: an observer can finally tell a rejected param from a query that actually threw. Nice.

One thing your own commit message put in my head. It says "helper function hasMongoOperator", but the function is living inside `handleMessage` as a closure, rebuilt on every message. `shared/helperFunctions.ts` is right there with `assertNever` in it. What changes, for you, if it moves there, both for the rebuilt-every-call bit and for being able to test the operator rule on its own without standing up a whole subscribe flow?

And while you're in that block: the three error replies (invalid params, queryFn threw, find threw) are the same `ws.send(...) + notifyObserver(...) + return` shape three times. Once the helper moves out, does that repetition start to bother you, or do you read it as fine where it is?

server/logic/publications.ts
```diff
@@ -65,6 +68,20 @@ export class Publications {
     }
   }
 
+  private sendPublicationError(
```
KevinOrfas, line 71: Ebuka, this is the move. `hasMongoOperator` reads as its own thing now, the test file pins the behaviour without needing a subscribe flow, and the three error replies you spotted collapse into one named call. The function reads top down and the noise is gone from `handleMessage`.

One question worth holding up to it.

Read `sendPublicationError` next to the branch a few lines above, where unknown publication name still does:

```ts
this.ws.send(clientId, {
  type: 'error',
  id: message.id,
  error: { code: 404, message: `Unknown publication: ${message.name}` },
});
this.notifyObserver(message, 'failed', 'unknown-publication');
return Promise.resolve();
```

Same three steps, same order. The only differences are `code: 404` and the message phrasing. So when the helper landed, four instances of the shape became three-plus-one rather than four-into-one. Which makes me curious where the line is for you between "same shape" and "same enough".

If the helper grows a `code` parameter and folds the unknown-publication branch in too, you get one function that owns "tell the client and observer this subscription failed", with reason and code as the things that vary. If it stays as-is, the helper is specifically about the 400-bracket failures, and unknown-publication sits outside on purpose. Either reading is defensible. The question is which boundary you think `sendPublicationError` is drawing.

A smaller thread while you're in it. The two `publication-query-failed` callers (queryFn threw, find threw) now share both the helper and the reason. An observer can tell "invalid params" from "query failed", but not "user code threw" from "mongo threw". Is that distinction worth carrying, or is "query failed" the right grain?

Iam-Ebuka: I just refactored the sendPublicationError function after review because what you pointed out is correct; all four error instances can use the function.

Secondly, I gave the observer a different reason for when mongo threw so that an observer can tell what exactly threw.

Iam-Ebuka (commented):
I just refactored the sendPublicationError function after review because what you pointed out is correct; all four error instances can use the function.

Secondly, I gave the observer a different reason for when mongo threw so that an observer can tell what exactly threw.

PR #52, opened 2026-06-02
Eko 279/3c4 gate late messages after stop

KevinOrfas:
Closes the gap left by 3.C.1's Definition of Done point 6: after `handle.stop()`, late `added` / `changed` / `removed` messages for a stopped subscription should not leak into the store.

- **problem:** `ConnectionManager.handleServerMessage` routes data into `store(collection)` with no liveness check. The only guard is the manager-wide `disposed` flag, so after a single `stop()` a late message for that collection still mutates the store.
- **test (red):** a late `changed` after `stop()` must not mutate the store (`tests/client/connectionManager.test.ts`). Currently `it.fails`, pinning the contract until the gate lands.
- **fix (green):** a per-collection liveness gate. `subscribe` marks the collection live, `stopSubscription` clears it once no live subscription remains, and `handleServerMessage` ignores data for a collection with no live subscription. The gate keys on collection, not subscription id, because `DataMsg` carries no id.

**A decision this forces.** 3.C.1 never had to say what happens to a message that arrives *after* `stop()`. Now we do, and two tests currently expect the opposite:

- the new test sends a late `changed` and expects the store **left alone** (the late message is ignored),
- the existing lifecycle test (`connectionManager.test.ts:161`) sends a late `removed` and expects the doc **deleted** (the late message is applied).

A single gate cannot do both: if it ignores the late `changed`, it ignores the late `removed` too.

This PR chooses **ignore every message after `stop()`**. Once a subscription is stopped, nothing more touches the store, and whatever was already cached stays put. It is the simpler rule and matches EkoLite having dropped the server-side merge box. The consequence: the lifecycle test changes so that after `stop()` the late `removed` no longer clears the doc, and the doc remains.

**Out of scope:** client-side eviction of a stopped subscription's docs, the server-side merge box / refcounted ownership (dropped by design, tracked as the `publications.refcount` gap), and reconnect.

EKO-279 · https://linear.app/ekohacks/issue/EKO-279/3c4-gate-late-messages-after-stop

### DIFF
--- a/client/clientSocket.ts
+++ b/client/clientSocket.ts
@@ -17,7 +17,7 @@ export function isServerMessage(data: unknown): data is ServerMessage {
 
   switch (type) {
     case 'ready':
-      return typeof msg.id === 'string';
+      return typeof msg.id === 'string' && typeof msg.collection === 'string';
     case 'result':
       return typeof msg.id === 'string';
     case 'added':
--- a/client/connectionManager.ts
+++ b/client/connectionManager.ts
@@ -213,14 +213,10 @@ export class ConnectionManager {
       case 'ready': {
         const subscription = this.subscriptions.get(message.id);
         if (subscription) {
-          // The server names the collection on ready. Fall back to the
-          // collection of the initial data buffered for this sub when an older
-          // server omits it.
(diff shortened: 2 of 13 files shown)


Iam-Ebuka (approved):
good stuff

PR #53, opened 2026-06-08
Eko 276/3d3 refactor websocketwrapper to nullable infrastructure

Iam-Ebuka:
- test: red - removes a client and notifies onDisconnect exactly once when it disconnects
- refactored websocket.ts

https://linear.app/ekohacks/issue/EKO-276/3d3-refactor-websocketwrapper-to-nullable-infrastructure

### DIFF
--- a/server/infrastructure/websocket.ts
+++ b/server/infrastructure/websocket.ts
@@ -12,12 +12,10 @@ interface ServerSocketLike {
   close(): void;
   onClose(cb: () => void): void;
   onMessage?(cb: (message: unknown) => void): void;
-  setClientId?(id: string): void; // For NullConnectionSource to set the stub's ID
-  setReceiveMessageHandler?(handler: (message: unknown) => void): void; // For NullConnectionSource to handle incoming messages
 }
 
 interface ConnectionSource {
-  onConnection(cb: (socket: ServerSocketLike) => unknown): void;
+  onConnection(cb: (socket: ServerSocketLike) => string): void;
   close(): Promise<void>;
   start?(): Promise<void>;
   attach?(fastify: FastifyInstance): Promise<void>;
@@ -46,9 +44,7 @@ export class WebSocketWrapper implements WebSocketInterface {
 
   private constructor(factory: ConnectionSourceFactory) {
     this.source = factory();
-    this.source.onConnection((socket) => {
-      this.handleConnection(socket);
-    });
+    this.source.onConnection((socket) => this.handleConnection(socket));
   }
 
   static create(): WebSocketWrapper {
@@ -96,14 +92,14 @@ export class WebSocketWrapper implements WebSocketInterface {
   }
 
   send(clientId: string, message: unknown): void {
-    const entry = this.clients.get(clientId);
-    if (!entry) {
-      return;
-    }
     if (message === undefined) {
       throw new Error('Cannot send undefined message to client');
     }
 
+    const entry = this.clients.get(clientId);
+    if (!entry) {
+      return;
+    }
     entry.socket.send(JSON.stringify(message));
   }
 
@@ -146,10 +142,9 @@ export class WebSocketWrapper implements WebSocketInterface {
     this.emitter.emit(MESSAGE_EVENT, { clientId, message });
   }
 
-  private handleConnection(socket: ServerSocketLike): void {
+  private handleConnection(socket: ServerSocketLike): string {
     const id = String(this.nextId++);
     this.clients.set(id, { socket });
-    socket.setClientId?.(id); // For NullConnectionSource: set the stub's ID
     socket.onMessage?.((message: unknown) => {
       this.receiveMessage(id, message);
     });
@@ -162,7 +157,7 @@ export class WebSocketWrapper implements WebSocketInterface {
       }
     });
 
-    // For NullConnectionSource: set the stub's ID and receiveMessageHandler
+    return id;
   }
 }
 
@@ -176,14 +171,14 @@ function isClientIdPayload(data: unknown): data is { clientId: string } {
 
 class WsConnectionSource implements ConnectionSource {
   private wss: WebSocketServer | null = null;
-  private listeners: ((s: ServerSocketLike) => unknown)[] = [];
+  private listeners: ((socket: ServerSocketLike) => string)[] = [];
   private port: number;
 
   constructor(port: number) {
     this.port = port;
   }
 
-  onConnection(cb: (socket: ServerSocketLike) => unknown): void {
+  onConnection(cb: (socket: ServerSocketLike) => string): void {
     this.listeners.push(cb);
   }
 
@@ -224,9 +219,9 @@ class WsConnectionSource implements ConnectionSource {
 
 class FastifyConnectionSource implements ConnectionSource {
   private fastify: FastifyInstance | null = null;
-  private listeners: ((s: ServerSocketLike) => unknown)[] = [];
+  private listeners: ((s: ServerSocketLike) => string)[] = [];
 
-  onConnection(cb: (socket: ServerSocketLike) => unknown): void {
+  onConnection(cb: (socket: ServerSocketLike) => string): void {
     this.listeners.push(cb);
   }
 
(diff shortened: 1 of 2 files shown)


KevinOrfas (commented):
Hi Ebuka, I sat with this one and ran tests against it rather than just reading the diff. Walking through it the way I read it.

**The heart of the change is right.** You took the test double knowledge out of `send` and `broadcast`. They used to stop and ask "is this a stub?" on every message. Now they just call `socket.send(...)` and let the null socket decide how to record it. That is what a nullable is for, so this is the direction we want. You did the same to disconnect: a client closing and the server closing now run through one `onClose` handler instead of the old separate `disconnect` method. I tested calling `close()` twice on a client and the disconnect still fires exactly once and the count still lands on zero, so the `if (this.clients.has(id))` guard is holding the line.

**Something nice happened, and it is worth understanding why.** You changed the null socket to keep `JSON.parse(JSON.stringify(message))`. That one line does something good on purpose. It records each message the way the wire would actually carry it, instead of holding onto the live object you handed in. I checked it by running the same two tests against your branch and against the commit before yours.

First, a message is now kept by value. Change the object after you send it and the recorded copy stays put:

```js
it('keeps an outbound message by value, so changing the object later does not change what was sent', () => {

  const payload = { type: 'hello', nested: { n: 1 } };
  ws.send(client.id, payload);
  payload.nested.n = 999;

  expect(client.messages).toEqual([{ type: 'hello', nested: { n: 1 } }]);
});
```

Passes on your branch. On the commit before yours it fails, because the old null held the live object, so the recorded copy quietly turned into `999`. You fixed that.

Second, a value that cannot cross the wire as is, like a `Date`, is now kept as the string a real socket would send:

```js
it('records a Date as the string a real socket would send', () => {

  ws.send(client.id, { type: 'tick', when: new Date('2026-06-09T00:00:00.000Z') });

  expect(client.messages).toEqual([{ type: 'tick', when: '2026-06-09T00:00:00.000Z' }]);
});
```

Same story. Passes on your branch, fails on the one before, where the old null kept a real `Date`. Both of these mean the null is now telling the truth about production, which is the whole point of it. I would add both to `websocket.test.ts` so nobody can quietly undo it later.

**Now the flip side of that same line.** The round trip that buys us all of the above is also the line that breaks when there is nothing to stringify. `JSON.stringify(undefined)` gives back `undefined`, not text, and then `JSON.parse(undefined)` throws:

```js
it('broadcast(undefined)', () => {

  ws.broadcast(undefined);

  expect(client.messages).toEqual([undefined]);
});
```

On your branch this throws `"undefined" is not valid JSON`. On the one before it passes, because the old null just stored `undefined`. So the same change that made the null honest about a `Date` made it throw on `undefined`. This is not a mistake to undo, it is a decision to make, and I will open a separate note with the two ways we could go so we can settle it together.

**The part I would most like you to rework.** The null knowledge did not leave, it moved into `handleConnection`. Every real connection, Fastify and `ws` included, now runs this:

```js
const nullSource = this.source as { lastCreatedStub?: StubbedClient | undefined };
if (nullSource.lastCreatedStub?.id === '') {
  nullSource.lastCreatedStub.setId(id);
  nullSource.lastCreatedStub.setReceiveMessageHandler((message) => this.receiveMessage(id, message));
  nullSource.lastCreatedStub = undefined;
}
```

Read it as a story. `handleConnection` is meant to handle any connection, but here it pauses to ask "are you secretly the null one, and have you got a stub waiting?" That is the same question we were glad to delete from `send` and `broadcast`. The empty string is standing in for "not wired up yet", and `lastCreatedStub` is a single slot that only works because one listener reads it at one exact moment. It does no harm at runtime, a real source has no `lastCreatedStub` so the line skips, and the Fastify and `ws` tests all pass. But it is weight on the shared path, and it is the smell we just cleaned up creeping back in.

The shape that solves it already lives in this file. Look at `WsConnectionSource` and `FastifyConnectionSource`: each one builds its own `ServerSocketLike` and hands it over, so `handleConnection` treats them all the same and never asks what kind they are. The null source can do the same if the socket carries one more optional step, say `setClientId?(id)`. Then `handleConnection` reads as one sentence for everyone:

```js
this.clients.set(id, { socket });
socket.setClientId?.(id);   // the null socket stamps its stub here; real sockets do not have this step
this.emitter.emit(CONNECTION_EVENT, { clientId: id });
```

The null source wires the stub inside its own socket, right next to `send` and `close`. The empty string check disappears, and the single slot disappears with it.

**One habit, not just this PR.** A few comments explain what the code is doing: `// UNIFIED: ...`, `// For NullConnectionSource: ...`, `id; // Mutable for testing`. We lean away from these because they drift and then mislead. That last one already does, since `id` is set by `handleConnection` in normal running, not by a test. Let the names and the structure carry the meaning and the code keeps telling the true story on its own. The `setClientId` shape above does most of that for free.

**To turn this into commits on this PR:**

- Run the three tests above against your branch, and against the commit before yours if you want to see the difference for yourself.
- Bring the value test and the `Date` test into `websocket.test.ts` as guards, so the wire fidelity you just gained cannot be quietly undone.
- Move the null wiring out of `handleConnection` using the `setClientId` shape.
- I will follow up with a separate note on the `undefined` question, and once we have picked a direction I will finish that last test with you.

The bones here are good and the core idea landed. Happy to pair on any of it.

server/infrastructure/websocket.ts
```diff
@@ -7,114 +7,204 @@ const CONNECTION_EVENT = 'connection';
 const DISCONNECTION_EVENT = 'disconnection';
 const MESSAGE_EVENT = 'message';
 
+interface ServerSocketLike {
+  send(data: string): void;
+  close(): void;
+  onClose(cb: () => void): void;
+  onMessage?(cb: (message: unknown) => void): void;
+  setClientId?(id: string): void; // For NullConnectionSource to set the stub's ID
+  setReceiveMessageHandler?(handler: (message: unknown) => void): void; // For NullConnectionSource to handle incoming messages
```
KevinOrfas: `setClientId?`, `setReceiveMessageHandler?` and `onMessage?` here only exist for the null socket. The real sources never implement them, so the shared contract is carrying the null source's private wiring. On the client side we kept seams like this off the shared interface on purpose: see `client/clientSocket.ts:110`, where `deliver` is an extra method on `NullWebSocket` with a comment saying it is explicitly not part of `WebSocketLike`. The only things the wrapper truly needs from any socket are `send`, `close`, `onClose`, and `onMessage` once we wire real inbound messages.

Also `setReceiveMessageHandler?` on line 16 is dead: `handleConnection` wires messages through `socket.onMessage?.()` (line 153), and nothing calls the interface member anywhere. Safe to delete.

server/infrastructure/websocket.ts
```diff
@@ -7,114 +7,204 @@ const CONNECTION_EVENT = 'connection';
...
 
   send(clientId: string, message: unknown): void {
-    this.server.send(clientId, message);
+    const entry = this.clients.get(clientId);
+    if (!entry) {
+      return;
+    }
+    if (message === undefined) {
+      throw new Error('Cannot send undefined message to client');
+    }
```
KevinOrfas: `send` and `broadcast` guard `undefined` in opposite orders. Here the early return for an unknown client (lines 100 to 102) runs before the `undefined` check, so `ws.send(aDisconnectedId, undefined)` swallows the bad payload silently, while `broadcast` (line 111) checks `undefined` first and always throws. Both new specs use a live client, so they pass either way and miss the gap. Move the `undefined` check above the client lookup so a bad payload is reported whether or not the client is still connected, and so the two methods behave the same.

server/infrastructure/websocket.ts
```diff
@@ -7,114 +7,204 @@ const CONNECTION_EVENT = 'connection';
...
+
+  private receiveMessage(clientId: string, message: unknown): void {
+    this.emitter.emit(MESSAGE_EVENT, { clientId, message });
+  }
+
+  private handleConnection(socket: ServerSocketLike): void {
+    const id = String(this.nextId++);
+    this.clients.set(id, { socket });
+    socket.setClientId?.(id); // For NullConnectionSource: set the stub's ID
+    socket.onMessage?.((message: unknown) => {
```
KevinOrfas, line 148: The real sources (`WsConnectionSource`, `FastifyConnectionSource`) never define `onMessage`, so this does nothing for them and `trackMessages` returns a live but never firing tracker on a real server. That matches the old behaviour (real inbound messages were never handled), so it is not a regression, but two things changed worth a line in the description: the old code threw on `trackMessages` for real instances to flag misuse, and now it quietly hands back an empty tracker. If real inbound messages are coming, this is the right seam; if not yet, a note saying so saves the next reader the confusion I had.

server/infrastructure/websocket.ts
```diff
@@ -7,114 +7,204 @@ const CONNECTION_EVENT = 'connection';
...
+    this.emitter.emit(CONNECTION_EVENT, { clientId: id });
+
+    socket.onClose(() => {
+      if (this.clients.has(id)) {
+        this.clients.delete(id);
+        this.emitter.emit(DISCONNECTION_EVENT, { clientId: id });
+      }
+    });
+
+    // For NullConnectionSource: set the stub's ID and receiveMessageHandler
```
KevinOrfas: This comment now sits at the end of `handleConnection` describing work that already ran above it (lines 152 to 153). Stale, safe to remove.

server/infrastructure/websocket.ts
```diff
@@ -130,210 +220,126 @@ class RealWebSocket implements WebSocketInterface {
...
-  constructor(id: string, server: StubbedWebSocket) {
+  readonly messages: unknown[];
+  id: string; // Mutable for testing, set by handleConnection
+  private closeHandler: (() => void) | undefined;
+  private receiveMessageHandler: ((message: unknown) => void) | undefined;
+
+  constructor(
+    messages: unknown[],
+    closeHandler?: () => void,
+    receiveMessageHandler?: (message: unknown) => void,
```
KevinOrfas: `id` lost its `readonly` and gained a public `setId` so it can be written after construction (the `lastCreatedStub` relay below). If we collapse that relay, this can go back to `readonly`. Separately, the third constructor parameter `receiveMessageHandler` on line 266 is never passed: `simulateConnection` calls `new StubbedClient(messages, ...)` with two args and sets the handler later. Drop the parameter.

server/infrastructure/websocket.ts
```diff
@@ -130,210 +220,126 @@ class RealWebSocket implements WebSocketInterface {
...
   }
 }
 
-class StubbedWebSocket implements WebSocketInterface {
-  private clients = new Map<string, StubbedClient>();
-  private emitter = new EventEmitter();
-  private nextId = 0;
+class NullConnectionSource implements ConnectionSource {
+  private listenersList: ((s: ServerSocketLike) => unknown)[] = [];
+  lastCreatedStub: StubbedClient | undefined;
```
KevinOrfas: This is the substantive one. To get the wrapper assigned id into the stub, the code stashes the stub here on `lastCreatedStub`, then `handleConnection` calls `socket.setClientId(id)` (line 152), which reaches back through `lastCreatedStub.setId(id)` (line 323) to mutate it after the fact. Same dance for the message handler (line 328). It holds together today only because this is synchronous and one connection is made at a time.

Compare `StubbedServer` in `client/clientSocket.ts:116`, which just takes its socket in the constructor (`private readonly socket`) and drives it. One reference, set once.

The honest difference is that here the wrapper assigns the id, so it does have to reach the test handle somehow. So a question rather than a prescription: could `StubbedClient` hold its null socket directly like `StubbedServer` does and read the id from it, so we drop `lastCreatedStub`, the `setClientId` hop and the mutable public id? If the stub and the null socket were one object implementing `ServerSocketLike`, `handleConnection` could set the id straight on it and the whole relay disappears. Keen to hear how you would surface the id without the back channel.

server/infrastructure/websocket.ts
```diff
@@ -130,210 +220,126 @@ class RealWebSocket implements WebSocketInterface {
...
+    const messages: unknown[] = [];
+    let onCloseCallback: (() => void) | undefined;
+
+    const socket: ServerSocketLike = {
+      send: (data: string) => {
+        // Null socket: record in-memory instead of sending over network
+        messages.push(JSON.parse(data));
+      },
+      close: () => {
+        // UNIFIED: call onClose callback like real socket does
```
KevinOrfas: A few comments read as notes to yourself mid refactor rather than to the next reader: this `// UNIFIED: ...`, plus `// Mutable for testing ...` (line 259) and `// Temporary ID ...` (line 268). They can all go; the code already says what it does.

tests/infrastructure/websocket.test.ts
```diff
@@ -70,4 +84,22 @@ describe('WebSocketWrapper (null)', () => {
...
+  it('it throws an error with a clear message if undefined is sent to clients', () => {
+    const ws = WebSocketWrapper.createNull();
+    const client = ws.simulateConnection();
+
+    expect(() => {
+      ws.send(client.id, undefined);
+    }).toThrow('Cannot send undefined message to client');
+  });
+
+  it('it throws an error with a clear message if undefined is broadcasted to clients', () => {
```
KevinOrfas: Both new specs start with the word `it` (`it('it throws an error ...')`), so the reporter prints 'it it throws ...'. Drop the leading `it` from the description. Tiny thing: the broadcast spec creates a connection on line 99 it does not need, since `broadcast` throws on `undefined` before it touches any client.

KevinOrfas (commented):
## Pushing the null seam down to the connection source: good direction

Thanks Ebuka. This moves the null seam to where it belongs. Before, the whole `WebSocketWrapper` was swapped for a parallel `StubbedWebSocket` that reimplemented client tracking, send, broadcast and disconnect, so the production logic was never the thing under test. Now the wrapper owns that logic for real and only the connection source varies. That is exactly the shape we want, and `trackConnections` / `trackDisconnections` now fire on the real sources too, where the old code used to throw.

So it works and it is tested. The inline notes are about the null wiring, which is where the cost of this design shows up later.

### The shape of it
The one substantive ask is the `lastCreatedStub` relay in `NullConnectionSource` (inline comment there). We solved this exact problem on the client side last week in `client/clientSocket.ts` (#47): the null class implements the socket interface cleanly, the test seam stays off the shared interface, and the test handle holds its socket by constructor reference. This PR takes a harder road to the same place, with null only methods on `ServerSocketLike` and an id written back into the stub after construction. Bringing it in line with the client side would drop a fair bit of the wiring and let `StubbedClient.id` go back to `readonly`.

The rest is small: one real consistency bug in the `undefined` guard ordering on `send`, a behaviour note on `trackMessages` for the real sources, a dead interface member and constructor parameter, and some refactor era comments to clear out. All inline.


Solid work overall. Happy to pair on the id question if it helps.

PR #55, opened 2026-06-11
Eko 277/3d4 refactor mongowrapper to nullable infrastructure

Iam-Ebuka:
**Refactor MongoWrapper to nullable infrastructure**

Next in the nullable infrastructure series, after the websocket and clientsocket refactors.

**What it does**

Moves `MongoWrapper` onto the nullable infrastructure pattern, so one wrapper backs both real Mongo and tests with no mocks. The wrapper now talks to a small `CollectionLike` seam, a real collection over the driver and a stubbed one for null instances. Both run through the same dispatch and mapping code, so the null instance behaves like the real thing rather than reimplementing it.

**What it unlocks**

`watchChanges` now works against real Mongo. It opens a change stream lazily on the first watcher, maps raw change events into our `ChangeEvent` shape for insert, update and remove, and closes the stream when the last watcher leaves. The id sits at the top level as `change.id`, with `_id` kept out of `fields`, matching the `ChangeEvent` contract.

**Knock on**

The publications subscribe handler reads the insert id from `change.id` and records live inserts in `documentIds`, so unsubscribe replays a `removed` for documents delivered through the live stream, not just the initial snapshot.

**Tooling**

Typecheck runs on push and in CI rather than on every commit, so commits stay fast while type errors are still caught before they land.

**Tests**

Null and integration coverage for the change stream path, kept in separate files.

Linear: https://linear.app/ekohacks/issue/EKO-277/3d4-refactor-mongowrapper-to-nullable-infrastructure

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,9 +25,10 @@ jobs:
 
             - name: "Install dependencies"
               run: npm ci
+            - name: "Type check"
+              run: npm run typecheck
             - name: "Lint code"
               run: npm run lint
             - name: "Run tests"
               run: npm test
-            - name: "Type check"
-              run: npm run typecheck
+            
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,3 +29,5 @@ jobs:
               run: npm run lint
             - name: "Run tests"
               run: npm test
+            - name: "Type check"
+              run: npm run typecheck
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,2 @@
 npx lint-staged
+npm run typecheck
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,2 +1,1 @@
-npm run typecheck
 npx lint-staged
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,2 +1,2 @@
-npx lint-staged
 npm run typecheck
+npx lint-staged
--- a/server/infrastructure/mongo.ts
+++ b/server/infrastructure/mongo.ts
@@ -1,104 +1,16 @@
-import { Db, MongoClient as Driver, ObjectId } from 'mongodb';
+import { Collection, MongoClient as Driver, ObjectId } from 'mongodb';
 import { ChangeEvent, isChangeEvent } from '../../shared/types.ts';
 import { ConfigurableResponse, EventEmitter, OutputTracker } from './outputTracker.ts';
 
-interface MongoInterface {
-  find<T>(collection: string, query: object): Promise<T[]>;
-  insert(collection: string, doc: object): Promise<void>;
-  update(collection: string, query: object, changes: object): Promise<void>;
-  remove(collection: string, query: object): Promise<void>;
-  watchChanges(collection: string, cb: (change: ChangeEvent) => void): () => void;
-  watcherCount(collection: string): number;
-  trackChanges(collection: string): OutputTracker;
+interface CollectionLike {
+  find(query: object): Promise<unknown[]>;
+  insertOne(doc: object): Promise<void>;
+  updateMany(query: object, changes: object): Promise<void>;
+  deleteMany(query: object): Promise<void>;
+  watch(onChange: (raw: unknown) => void): () => void;
 }
 
-export class MongoWrapper {
-  private client: MongoInterface;
-
-  private constructor(client: MongoInterface) {
-    this.client = client;
-  }
-
-  static create(uri: string): MongoWrapper {
-    return new MongoWrapper(new RealMongo(uri));
-  }
-
-  static createNull(
-    options: {
-      find?: unknown[];
-      insert?: unknown[];
-      update?: unknown[];
-      remove?: unknown[];
-    } = {},
-  ): MongoWrapper {
-    return new MongoWrapper(new StubbedMongo(options));
-  }
-
-  async find<T>(collection: string, query: object): Promise<T[]> {
-    return this.client.find<T>(collection, query);
-  }
-
-  async insert(collection: string, doc: object): Promise<void> {
-    return this.client.insert(collection, doc);
-  }
-
-  async update(collection: string, query: object, changes: object): Promise<void> {
-    return this.client.update(collection, query, changes);
-  }
-
-  async remove(collection: string, query: object): Promise<void> {
-    return this.client.remove(collection, query);
-  }
-
-  watchChanges(collection: string, cb: (change: ChangeEvent) => void): () => void {
-    return this.client.watchChanges(collection, cb);
-  }
-
-  watcherCount(collection: string): number {
-    return this.client.watcherCount(collection);
-  }
-
-  trackChanges(collection: string): OutputTracker {
-    return this.client.trackChanges(collection);
-  }
-}
-
-class RealMongo implements MongoInterface {
-  private db: Db;
-
-  constructor(uri: string) {
-    const client = new Driver(uri);
-    this.db = client.db();
-  }
-
-  async find<T>(collection: string, query: object): Promise<T[]> {
-    return (await this.db.collection(collection).find(query).toArray()) as T[];
-  }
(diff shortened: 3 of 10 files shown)


server/infrastructure/mongo.ts
```diff
@@ -1,120 +1,150 @@
...
-interface MongoInterface {
-  find<T>(collection: string, query: object): Promise<T[]>;
-  insert(collection: string, doc: object): Promise<void>;
-  update(collection: string, query: object, changes: object): Promise<void>;
-  remove(collection: string, query: object): Promise<void>;
-  watchChanges(collection: string, cb: (change: ChangeEvent) => void): () => void;
-  watcherCount(collection: string): number;
-  trackChanges(collection: string): OutputTracker;
+interface CollectionLike {
+  find(query: object): Promise<unknown[]>;
```
KevinOrfas: We should keep the generic `T`

.github/workflows/ci.yml
```diff
@@ -29,3 +29,5 @@ jobs:
               run: npm run lint
             - name: "Run tests"
               run: npm test
+            - name: "Type check"
```
KevinOrfas: 1. types first 
2. lint 
3. tests

.husky/pre-commit
```diff
@@ -1 +1,2 @@
 npx lint-staged
+npm run typecheck
```
KevinOrfas: I feel that types should be at the top, types first then linter

.github/workflows/ci.yml
Iam-Ebuka: alright

.husky/pre-commit
Iam-Ebuka: alright

Iam-Ebuka: done

.github/workflows/ci.yml
Iam-Ebuka: done

server/logic/publications.ts
```diff
@@ -176,8 +176,8 @@ export class Publications {
 
       const cleanup = this.mongo.watchChanges(collection, (change: ChangeEvent) => {
         if (change.type === 'insert') {
-          documentIds.add(change.fields._id as string);
-          this.ws.send(clientId, addedMessage(collection, change.fields));
+          documentIds.add(change.id);
+          this.ws.send(clientId, addedMessage(collection, { _id: change.id, ...change.fields }));
```
KevinOrfas, line 180: For a live insert the id sits on `change.id`, not inside `change.fields`. The ChangeEvent contract keeps `_id` out of `fields` (extractFields strips it in mongo.ts), so the old `change.fields._id` was always undefined and we tracked an undefined id. Rebuilding `{ _id: change.id, ...change.fields }` keeps this identical to the initial find loop just above, and `addedMessage` reads `_id` back out. The new unsubscribe test pins it.

server/infrastructure/mongo.ts
```diff
@@ -129,80 +165,181 @@ class StubbedMongo implements MongoInterface {
...
+    const fullDocument = (raw as { fullDocument?: unknown }).fullDocument;
+    const updateDescription = (raw as { updateDescription?: unknown }).updateDescription as
+      | { updatedFields?: Record<string, unknown> }
+      | undefined;
+    const id = getId(documentKey?._id ?? (fullDocument as { _id?: unknown })._id);
+    return {
+      type: 'update',
+      collection: typeof collection === 'string' ? collection : '',
+      id,
+      fields: fullDocument ? extractFields(fullDocument) : (updateDescription?.updatedFields ?? {}),
```
KevinOrfas, line 301: This needs to stay a ternary rather than `extractFields(fullDocument) || updatedFields || {}`. extractFields always returns an object and `{}` is truthy, so the `||` chain could never reach updatedFields, and a missing fullDocument would send empty fields instead of the change stream delta. The ternary falls back to updatedFields only when fullDocument is genuinely absent, say a doc updated then deleted before updateLookup resolves.

server/infrastructure/mongo.ts
```diff
@@ -1,120 +1,156 @@
...
-  find?: unknown[];
-  insert?: unknown[];
-  update?: unknown[];
-  remove?: unknown[];
+  watch(onChange: (raw: unknown) => void): () => void {
+    const changeStream = this.collection.watch([], { fullDocument: 'updateLookup' });
+    changeStream.on('change', onChange);
+    changeStream.on('error', (err) => {
+      // Ignore stream errors here. The caller can still report errors through regular driver behavior.
+      console.error(`Mongo change stream error on ${this.collection.collectionName}`, err);
```
KevinOrfas, line 138: Logging here beats the old empty handler that swallowed the error silently. One tidy though: the comment just above still says we ignore the error, which now reads against the console.error. Worth rewording so the comment matches.

server/infrastructure/mongo.ts
```diff
@@ -129,80 +165,181 @@ class StubbedMongo implements MongoInterface {
...
+    return null;
+  }
+
+  const operationType = (raw as { operationType?: unknown }).operationType;
+  const collection = (raw as { ns?: { coll?: unknown } }).ns?.coll;
+  const documentKey = (raw as { documentKey?: { _id?: unknown } }).documentKey;
+
+  if (operationType === 'insert') {
+    const fullDocument = (raw as { fullDocument?: unknown }).fullDocument;
+    const id = getId(documentKey?._id ?? (fullDocument as { _id?: unknown })._id);
```
KevinOrfas: Soft spot I want us to look at when we pair. The cast says fullDocument is a non null object, so the `?.` got dropped here and on the update branch below. But fullDocument really can be undefined, and the update branch even guards it two lines down. This only holds because a real change event always carries documentKey._id, so this fallback never actually runs. Do we guard it properly or drop the dead fallback, rather than leave the cast papering over a nullable?

server/infrastructure/mongo.ts
```diff
@@ -1,120 +1,156 @@
...
   }
+}
 
-  async remove(collection: string, query: object): Promise<void> {
-    await this.db.collection(collection).deleteMany(query);
+class RealCollection implements CollectionLike {
+  constructor(private readonly collection: Collection) {}
+
+  find<T>(query: object): Promise<T[]> {
+    return this.collection.find(query).toArray() as Promise<T[]>;
```
KevinOrfas, line 117: Threading the generic through CollectionLike.find means the only cast sits here at the real driver boundary, and the public find<T> on the wrapper no longer needs its own. Same behaviour at runtime, cleaner types.

KevinOrfas (commented):
On the hooks and CI: pre-commit now runs lint-staged only, pre-push runs typecheck then test:changed, and CI runs typecheck first before lint and tests. That keeps commits fast, since the whole project tsc is too slow to sit on every commit, moves the heavier gates to push, and gives the quickest failure signal in CI. Happy with the split.

server/logic/publications.ts
Iam-Ebuka: oh, yes. good catch

server/infrastructure/mongo.ts
Iam-Ebuka: Yes I understand this now

server/infrastructure/mongo.ts
Iam-Ebuka: nice, I understand this

server/infrastructure/mongo.ts
KevinOrfas: Applied B: both the insert and update branches now use `const id = getId(documentKey?._id)`, no cast, no fullDocument fallback (5d9397b).

Reason: the `as { _id?: unknown }` cast asserted a non null object, which is what let lint strip the `?.` and remove the runtime guard. fullDocument is really `unknown` and can be undefined, so the assertion was false. documentKey is on every change event so the fallback never ran anyway, and getId already returns `''` for a missing id, so there was nothing left to guard. Dropping the cast removes the whole problem.

Iam-Ebuka: I think we should guard properly

PR #61, opened 2026-06-15
Eko 278/3d5 refactor filestoragewrapper and scriptrunnerwrapper to nullable

Iam-Ebuka:
- refactor of fileStorageWrapper to nullable structure.
- refactor of scriptRunnerWrapper to nullable structure.

https://linear.app/ekohacks/issue/EKO-278/3d5-refactor-filestoragewrapper-and-scriptrunnerwrapper-to-nullable


KevinOrfas (commented):
Good direction. All four wrappers look the same now: the wrapper holds the emitter, the adapter has a `watch()`, like Mongo and the socket. The notes below are mostly about recording getting switched on in the wrong places, plus one design question about the real path.

On tests: this PR adds two new tests, but they mostly repeat ones we already have. The new file storage test repeats the existing `tracks save, exists, and remove operations` test. The new script runner test repeats the existing `tracks successful executions` test. The thing this PR actually changed, real instances now sending events, has no test at all. That is the test worth adding, once you have settled the real path question above.

server/infrastructure/fileStorage.ts
```diff
@@ -1,98 +1,144 @@
...
+    return new FileStorageWrapper(factory.instance());
   }
 
   async save(name: string, data: Buffer): Promise<void> {
     if (!name) {
       throw new Error('File name cannot be empty');
     }
-    return this.fs.save(name, data);
+    await this.adapter.writeFile(name, data);
+    this.openWatchIfNeeded();
```
KevinOrfas: `trackChanges()` is what turns on recording of file operations. The trouble is that `save` turns it on only after it has already run and sent its event, so that event is missed. I checked it. Call `save`, then `trackChanges`, and nothing is recorded. `exists` and `remove` turn it on before their work, so the three do not even match each other.

The fix is to let only `trackChanges` turn recording on, which it already does, and delete `openWatchIfNeeded()` from `save`, `exists`, `remove`, and `exec`. Nothing that gets recorded changes. Mongo does it the same way. The empty `return;` just below can go too. (Same shape in `exec` at `scriptRunner.ts:34`.)

server/infrastructure/fileStorage.ts
```diff
@@ -1,98 +1,144 @@
...
+      this.emitter.off(CHANGE_EVENT, listener);
+    };
   }
 }
 
-interface StubbedFileSystemOptions {
-  save?: unknown[];
-  exists?: Error[];
-  remove?: unknown[];
+class StubbedFileSystemFactory {
```
KevinOrfas: This factory does not do anything. `instanceCache` is always `undefined`, so the `if` never runs and `instance()` just builds a new `StubbedFileStorage` every time, which is what `createNull` did directly before this change.

So delete the class and go back to `new FileStorageWrapper(new StubbedFileStorage(options))`. Mongo has a factory because it caches one per collection name. Here there is nothing to cache.

On the type: `?: T | undefined` is redundant, since the `?` already means 'or undefined', but it goes away with the class anyway.

server/infrastructure/fileStorage.ts
```diff
@@ -1,98 +1,144 @@
...
   constructor(basePath: string) {
     this.basePath = basePath;
   }
 
-  async save(name: string, data: Buffer): Promise<void> {
+  async writeFile(name: string, data: Buffer): Promise<void> {
     const fullPath = this.resolve(name);
     await mkdir(dirname(fullPath), { recursive: true });
     await writeFile(fullPath, data);
+    this.emitter.emit(CHANGE_EVENT, { type: 'save', name, data });
```
KevinOrfas, line 105: Real `save`, `exists`, `remove`, and `exec` now build and send an event every time they run. They did not before. `trackChanges` used to just throw on a real instance. Since `trackChanges` is only used by tests, in production nobody reads these events, so it is a little wasted work on every file operation.

Did you mean to make the real path recordable, or should real do nothing on `watch()` and leave recording to the null version? Your call. (`RealProcessRunner.exec` at `scriptRunner.ts:64` does the same.)

server/infrastructure/fileStorage.ts
```diff
@@ -1,98 +1,144 @@
...
-    return this.fs.trackChanges();
+    this.openWatchIfNeeded();
+    return new OutputTracker(this.emitter, CHANGE_EVENT);
+  }
+
+  private openWatchIfNeeded(): void {
+    if (this.stopWatch) {
+      return;
+    }
+    this.stopWatch = this.adapter.watch((raw) => {
```
KevinOrfas, line 78: `stopWatch` is saved but never called, so there is no way to turn the listener off. Harmless now, since these run for the whole life of the app, but Mongo has a `closeWatchIfUnused` to match if it ever matters.

server/infrastructure/fileStorage.ts
Iam-Ebuka: I didn't see where you are referring to with this: ` On the type: ?: T | undefined is redundant, since the ? already means 'or undefined', but it goes away with the class anyway.`

server/infrastructure/fileStorage.ts
Iam-Ebuka: Yeah, that was the idea of the refactor. The real should do the same thing as the nullable but we can pair on this to decide what shape this takes

server/infrastructure/fileStorage.ts
Iam-Ebuka: Can we pair on this one?

KevinOrfas (approved):

PR #75, opened 2026-06-23
Eko 219/4a2 structured error for unknown methods

Iam-Ebuka:
- test: throws structured error for unknown method
- Add structured 404 errors for unknown methods.

Replaces runtime failures with a consistent RpcError when calling undefined methods, adds test coverage, and centralizes error creation via methodNotFound().

https://linear.app/ekohacks/issue/EKO-219/4a2-structured-error-for-unknown-methods


shared/types.ts
```diff
@@ -63,3 +63,17 @@ export function isChangeEvent(data: unknown): data is ChangeEvent {
 // ── Method definitions ──────────────────────────────────────────────────────
 
 export type MethodFn = (...args: unknown[]) => Promise<unknown>;
+
+export function methodNotFound(code: number, name: string): never {
```
KevinOrfas: Have the factory build and return the error rather than throw inside it, and bake in the 404 so the caller doesn't pass it (nothing stops `methodNotFound(200, name)` today, and 404 is the whole point of a not-found helper):

```ts
export function methodNotFound(name: string): RpcError {
  return new RpcError(404, `Method not found: ${name}`);
}
```

Then throw at the call site in `methods.ts`. You'll need to `export` `RpcError` for the return type.

shared/types.ts
```diff
@@ -63,3 +63,17 @@ export function isChangeEvent(data: unknown): data is ChangeEvent {
...
+  throw new RpcError(code, `Method not found: ${name}`);
+}
+
+class RpcError extends Error {
```
KevinOrfas: `EkoLiteError` already exists in `shared/protocol.ts` (`{ code, message, details? }`) and it's what `Publications` returns. `class RpcError extends Error implements EkoLiteError` keeps the two error types the same shape, so 4.B can treat method and publication errors together. Name's good, keep it.

server/logic/methods.ts
```diff
@@ -13,8 +13,9 @@ export class Methods {
   async call(name: string, args: unknown[]): Promise<unknown> {
     const method = this.methods.get(name);
     if (!method) {
-      throw new Error(`Method not found: ${name}`);
+      methodNotFound(404, name);
```
KevinOrfas: This throws but doesn't read like it (it only compiles because the helper is typed `never`). With the factory returning the error, make the throw explicit:

```ts
if (!method) {
  throw methodNotFound(name);
}
```

tests/logic/methods.test.ts
```diff
@@ -9,6 +9,14 @@ describe('Methods', () => {
...
 
+  it('throws structured error for unknown method', async () => {
+    const methods = new Methods();
+    await expect(methods.call('nope', [])).rejects.toMatchObject({
+      code: 404,
+      message: 'Method not found: nope',
+    });
+  });
+
   it('throws when calling a method that is not defined', async () => {
```
KevinOrfas: This only checks the message, and the structured test just above covers the same path with more. Drop this one.

KevinOrfas (commented):
This works and CI's green, and putting `methodNotFound` in `shared/types.ts` is the right move since 4.B will want the same factory. A few inline notes.

One question: this branch carries the duplicate-define guard and the redefine and non-string tests from #40. Are the two going up together, or does #40 merge first and this rebase on top?

PR #81, opened 2026-06-25
Eko 216/4b2 return error for failed methods

Iam-Ebuka:
### Summary

Adds RPC error handling for method execution failures and missing methods.

### Changes
Wrapped Methods.call() in RpcHandler.handleMessage() with a try/catch.
Returns an RPC error response (type: 'error') instead of allowing exceptions to escape the handler.
Preserves the original request id in all error responses so callers can correlate failures with requests.
Normalizes all thrown values to the EkoLiteError shape:
Existing RpcError instances (e.g. method not found) are passed through.
Unexpected errors are converted to { code: 500, message }.

### Added tests covering:
Missing method (404 Method not found)
Method implementation throwing an error (500 normalized error)
Result

RPC callers now always receive a response for a method invocation, either:

`{ type: 'result', id, result }`

or

`{ type: 'error', id, error }`

with a consistent EkoLiteError payload shape across RPC and publication error handling.

https://linear.app/ekohacks/issue/EKO-216/4b2-return-error-for-failed-methods

### DIFF
--- a/client/connectionManager.ts
+++ b/client/connectionManager.ts
@@ -1,6 +1,12 @@
 import { ClientSocketWrapper } from './clientSocket.ts';
 import { ReactiveStore } from './reactiveStore.ts';
-import { DataMsg, ServerMessage, SubscribeMsg, UnsubscribeMsg } from '../shared/protocol.ts';
+import {
+  DataMsg,
+  MethodMsg,
+  ServerMessage,
+  SubscribeMsg,
+  UnsubscribeMsg,
+} from '../shared/protocol.ts';
 import { assertNever } from '../shared/helperFunctions.ts';
 
 export interface SubscriptionHandle {
@@ -36,6 +42,11 @@ const generateSubscriptionId = (): string => {
   return globalThis.crypto.randomUUID();
 };
 
+interface PendingRequest {
+  resolve: (value: unknown) => void;
+  reject: (error: unknown) => void;
+}
+
 export class ConnectionManager {
   private readonly socket: ClientSocketWrapper;
   private readonly stores = new Map<string, ReactiveStore>();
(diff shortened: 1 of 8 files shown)


KevinOrfas (changes requested):
Ebuka, this closes the loop we opened in 4.B.1. Every call now comes back as either a result or an error, the `id` rides along so the caller can match the response to its request, and the error payload matches what Publications already sends, so the client gets one shape to read. The try/catch landed right where it needed to. Three things to talk through, one I'd want in before it merges.

**1. The normaliser catches more than our own errors.** In `shared/types.ts` the guard is `error instanceof Error && 'code' in error`. The question it's really trying to ask is "is this one of ours", but `'code' in error` is wider than that. Node's IO and network errors are genuine `Error` instances that also carry a `code`, except theirs is a string like `ENOENT` or `ECONNREFUSED`. One of those thrown from inside a method would flow straight through as `{ code: 'ENOENT', message }`, and `EkoLiteError` promises `code` is a number. The `as RpcError` cast is what keeps the compiler quiet about it.

`RpcError` is declared just above in the same file, so we can ask the precise question and drop the cast at the same time:

```ts
export function rpcHandleMessageError(error: unknown): EkoLiteError {
  if (error instanceof RpcError) {
    return { code: error.code, message: error.message };
  }
  return {
    code: 500,
    message: error instanceof Error ? error.message : String(error),
  };
}
```

Anything that isn't ours then collapses to a clean 500, which is exactly what your 500 test already proves. Worth settling here, since this is the contract the client caller reads next.

**2. The factory name reads like a class.** `RpcHandleMessageError` with the capital R looks like a constructor, but it's a factory that returns a value. Its neighbour `methodNotFound` is camelCase and names what it builds rather than where it runs. Something like `toEkoLiteError` or `normalizeRpcError` would line up with it. Does `toEkoLiteError` read right to you?

**3. One test title claims more than it checks.** `'throws a structured 404, or the method body itself throws'` names two cases, but the body only exercises the unknown method 404. The body-throws case is the 500 test below it. Trimming the title to just the 404 keeps it honest about what it asserts.

Points 2 and 3 are quick. Point 1 is the one that matters: it's a real gap between what the type promises and what can land in it.

KevinOrfas (approved):

KevinOrfas (approved):

PR #82, opened 2026-06-25
Eko 214/4c1 call sends message and resolves promise

Iam-Ebuka:
### Summary

Adds client-side RPC support to ConnectionManager, enabling request/response method calls over the WebSocket connection.

### Changes
Added call(name, ...args) to send method messages and return a promise for the response.
Added tracking of pending RPC requests keyed by request ID.
Implemented handling of result messages to resolve pending method calls and clean up request state.
Extended error message handling to reject pending method calls with the received EkoLiteError.
Refactored ID generation into a shared utility used by both subscriptions and RPC calls.

### Tested
Verifies call() sends a method message with the expected name, params, and generated ID.
Verifies a matching result message resolves the returned promise with the server result.
Verifies a matching error message rejects the returned promise with the server-provided EkoLiteError.

### Notes
No timeout or retry behavior is included in this change. Pending RPC calls remain unresolved until a matching result or error message is received

https://linear.app/ekohacks/issue/EKO-214/4c1-call-sends-message-and-resolves-promise

### DIFF
--- a/client/connectionManager.ts
+++ b/client/connectionManager.ts
@@ -1,6 +1,12 @@
 import { ClientSocketWrapper } from './clientSocket.ts';
 import { ReactiveStore } from './reactiveStore.ts';
-import { DataMsg, ServerMessage, SubscribeMsg, UnsubscribeMsg } from '../shared/protocol.ts';
+import {
+  DataMsg,
+  MethodMsg,
+  ServerMessage,
+  SubscribeMsg,
+  UnsubscribeMsg,
+} from '../shared/protocol.ts';
 import { assertNever } from '../shared/helperFunctions.ts';
 
 export interface SubscriptionHandle {
@@ -36,6 +42,10 @@ const generateSubscriptionId = (): string => {
   return globalThis.crypto.randomUUID();
 };
 
+interface PendingRequest {
+  resolve: (value: unknown) => void;
+}
+
 export class ConnectionManager {
   private readonly socket: ClientSocketWrapper;
   private readonly stores = new Map<string, ReactiveStore>();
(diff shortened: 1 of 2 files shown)


KevinOrfas (changes requested):
Ebuka, the request and response wiring is right: each call gets a UUID, the pending map keys on it, and both `result` and `error` find their caller and clean up after themselves. Checking the request map before subscriptions in the `error` case is the correct order. Two gaps though, both on the paths that never get a happy reply, and both already have a worked answer elsewhere in this file.

**1. A failed send leaves the caller waiting forever.** In `call()` the send catch deletes the pending entry but never settles the promise:

```ts
this.socket.send(message).catch(() => {
  this.pendingRequests.delete(id);
});
```

So when the socket rejects the send, the map gets tidied but the promise you handed back stays pending for good. `subscribe()` a few lines up already has the shape we want, it deletes and then rejects:

```ts
this.socket.send(message).catch((error: unknown) => {
  const request = this.pendingRequests.get(id);
  if (request) {
    request.reject(error);
    this.pendingRequests.delete(id);
  }
});
```

**2. Disposal drops calls that are still in flight.** `dispose()` walks the subscriptions and rejects each one through `stopSubscription`, but it never touches `pendingRequests`. Since `dispose()` also runs from the socket's `onClose`, a connection that drops while a call is open leaves every waiting `call()` promise hanging, with no way to retry because the manager is now disposed. Same move as subscriptions, settle them on the way out:

```ts
for (const request of this.pendingRequests.values()) {
  request.reject(new Error('connection disposed'));
}
this.pendingRequests.clear();
```

Both of these want a test next to them, the same way the result and error paths got one.

**3. The id helper is named for one of its two callers now.** The summary says id generation moved to a shared utility, but `call()` reuses `generateSubscriptionId` as is, so an RPC request id now comes from a function named for subscriptions. A neutral name like `generateId` would match what it actually does now.

**4. Small one:** the test's `describe('ClientSocketWrapper call', ...)` is exercising `ConnectionManager.call`, not the socket wrapper, so the suite name points at the wrong unit.

Points 1 and 2 are the ones I'd hold the merge on. They're the same lesson really: the happy reply settles the promise, but the failure and teardown paths have to settle it too, or the caller waits forever.

KevinOrfas (approved):

PR #83, opened 2026-06-25
Eko 211/5a1 accept bam files

Iam-Ebuka:
### Summary

Introduces configurable file extension validation for uploads.

### Changes
Added validate(name) to Files to check if a file extension is allowed before upload.
Default allowed extension is ['bam'].
Allowed extensions are now configurable via constructor options.
Updated validation to support multiple formats without code changes.
Normalized comparison to be case-insensitive (e.g. SAMPLE.BAM is valid).

### Implementation
Extracts file extension using node:path.extname.
Compares against allowedExtensions set stored per instance.
Defaults to ['bam'] when no configuration is provided.

### Testing
Verifies default behavior accepts .bam and rejects unsupported extensions.
Verifies custom configuration allows additional formats (e.g. .cram).
Ensures validation works independently of storage and database layers.

### Notes
This change only validates file names; it does not block or persist uploads.
Upload enforcement is handled separately in a later step (5.A.2).

https://linear.app/ekohacks/issue/EKO-211/5a1-accept-bam-files

### DIFF
--- a/client/connectionManager.ts
+++ b/client/connectionManager.ts
@@ -1,6 +1,12 @@
 import { ClientSocketWrapper } from './clientSocket.ts';
 import { ReactiveStore } from './reactiveStore.ts';
-import { DataMsg, ServerMessage, SubscribeMsg, UnsubscribeMsg } from '../shared/protocol.ts';
+import {
+  DataMsg,
+  MethodMsg,
+  ServerMessage,
+  SubscribeMsg,
+  UnsubscribeMsg,
+} from '../shared/protocol.ts';
 import { assertNever } from '../shared/helperFunctions.ts';
 
 export interface SubscriptionHandle {
@@ -36,6 +42,10 @@ const generateSubscriptionId = (): string => {
   return globalThis.crypto.randomUUID();
 };
 
+interface PendingRequest {
+  resolve: (value: unknown) => void;
+}
+
 export class ConnectionManager {
   private readonly socket: ClientSocketWrapper;
   private readonly stores = new Map<string, ReactiveStore>();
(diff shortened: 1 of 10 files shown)


KevinOrfas (changes requested):
This does what 5.A.1 asks: an extension allow list with a sensible `['bam']` default and a configurable override, case folded so `SAMPLE.BAM` passes. Holding the actual upload enforcement back for 5.A.2 reads right. Two small things to tidy before it lands.

**1. The field is misspelled.** `allowedExtenstions` (the private field at the top of `Files`, set in the constructor and read in `validate`) has an extra `t`. The option and the `FilesOptions` interface are both spelled correctly as `allowedExtensions`, so it's only the private field that's off. Worth fixing while it's new, since it'll autocomplete wrong from here on.

**2. The extension is now pulled out in two places, slightly differently.** `buildStoredFile` already does `extname(input.name).replace(/^\./, '')` when it records the stored document, and `validate` does the same plus `.toLowerCase()`. So the "name to extension" step lives twice, and the two don't agree on case: an upload of `SAMPLE.BAM` gets stored with extension `BAM` but validated as `bam`. A small shared helper, say `extensionOf(name)` returning the lowercased extension and used by both, keeps them from drifting once `validate` gets wired into `upload` in 5.A.2. Do you want storage to fold case the same way, or keep the original for display?

One tiny extra, not a blocker: the test title `'validates files by the extention'` has a typo (extention), and `'allows set configuration rather than a constant inside the check'` describes the implementation more than the behaviour. Something like `'accepts a configured extension such as .cram'` reads clearer.

KevinOrfas (approved):

PR #84, opened 2026-06-25
Eko 210/5a2 reject non bam files

Iam-Ebuka:
### Summary

Adds validation guard to file uploads to prevent unsupported file types from being persisted.

### Changes
Enforces validate() inside Files.upload() before any storage or database writes.
Rejects uploads with unsupported extensions using a 400 error.
Ensures invalid uploads do not write bytes to disk or create database records.
Standardises upload errors to the shared EkoLiteError shape ({ code, message }) for consistency across the system.

### Behaviour
Allowed files (e.g. .bam) proceed normally through storage and persistence.

**Disallowed files (e.g. .txt) are rejected early with:**

`{ code: 400, message: 'Unsupported file type: .txt' }`
No side effects occur for rejected uploads.

### Testing
Verifies rejected uploads do not create files on disk.
Verifies upload() throws a 400 error for unsupported file types.
Ensures validation happens before any persistence logic.

### Notes
This only enforces domain validation at the logic layer.
HTTP response mapping and client-side handling are handled separately in later work.

https://linear.app/ekohacks/issue/EKO-210/5a2-reject-non-bam-files

### DIFF
--- a/client/connectionManager.ts
+++ b/client/connectionManager.ts
@@ -1,6 +1,12 @@
 import { ClientSocketWrapper } from './clientSocket.ts';
 import { ReactiveStore } from './reactiveStore.ts';
-import { DataMsg, ServerMessage, SubscribeMsg, UnsubscribeMsg } from '../shared/protocol.ts';
+import {
+  DataMsg,
+  MethodMsg,
+  ServerMessage,
+  SubscribeMsg,
+  UnsubscribeMsg,
+} from '../shared/protocol.ts';
 import { assertNever } from '../shared/helperFunctions.ts';
 
 export interface SubscriptionHandle {
@@ -36,6 +42,10 @@ const generateSubscriptionId = (): string => {
   return globalThis.crypto.randomUUID();
 };
 
+interface PendingRequest {
+  resolve: (value: unknown) => void;
+}
+
 export class ConnectionManager {
   private readonly socket: ClientSocketWrapper;
   private readonly stores = new Map<string, ReactiveStore>();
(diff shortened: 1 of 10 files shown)


KevinOrfas (changes requested):
WIP

KevinOrfas (changes requested):
**1. Duplicate test.** `tests/logic/files.test.ts` now has `'accepts a configured extension such as .cram'` twice, once near the top of the validate block and once lower down. That test already lives on main from #83, so the rebase re-added a copy of it. Drop the extra one, so the only test this PR adds is the upload guard one.

**2. Still open from the last pass:**
- `upload` computes `extensionOf(input.name)` and then calls `validate(input.name)`, which parses it again. Compute it once and pass it in. Just a double parse per upload.
- The test name `'upload questions files first before saving'` reads oddly. Something like `'rejects an unsupported upload before writing anything'` says what it checks.
- The body says no database record is created for a rejected upload, but the test only asserts the disk side (`storage.exists`). A quick assert that the collection is still empty would pin that claim.

KevinOrfas (approved):
Applied the review directly on this branch in dccdc8d, so flagging that here for transparency: dropped the duplicate 'accepts a configured extension' test, parse the extension once and share the allow-list check via a private isAllowed, renamed the guard test, and added a no-insert assertion so a rejected upload is pinned on both disk and database. Full suite, lint and typecheck green locally. Good to merge.

KevinOrfas (approved):

PR #86, opened 2026-06-25
Eko 285/5a3 route returns 400 for invalid files

Iam-Ebuka:
## Summary

This PR updates the file upload route to correctly return validation errors from `Files.upload` as HTTP responses instead of allowing them to surface as generic 500 errors.

### Changes

* Added an integration test covering uploads of unsupported file types (`.txt`).
* Updated `POST /api/files` to normalize upload errors using `toEkoLiteError`.
* Added centralized Fastify error handling so routes can throw and rely on a single error-to-response mapping.
* Standardized API error responses to the shared `{ code, message }` format.

#### Invalid file upload

Before:

* Unsupported files could result in a generic 500 response.
* Error responses were not consistently structured.

After:

* Unsupported files return `400` with an `EkoLiteError` response body.
* No file is stored when validation fails.

#### Valid file upload

Behavior remains unchanged:

* Returns `201 { id, name }`.

### Testing

Added integration coverage to verify:

* Uploading an unsupported file returns `400`.
* The response contains the expected `EkoLiteError` structure.
* No file is persisted when validation fails.
* Valid uploads continue to return `201`.

https://linear.app/ekohacks/issue/EKO-285/5a3-route-returns-400-for-invalid-files

### DIFF
--- a/client/connectionManager.ts
+++ b/client/connectionManager.ts
@@ -1,6 +1,12 @@
 import { ClientSocketWrapper } from './clientSocket.ts';
 import { ReactiveStore } from './reactiveStore.ts';
-import { DataMsg, ServerMessage, SubscribeMsg, UnsubscribeMsg } from '../shared/protocol.ts';
+import {
+  DataMsg,
+  MethodMsg,
+  ServerMessage,
+  SubscribeMsg,
+  UnsubscribeMsg,
+} from '../shared/protocol.ts';
 import { assertNever } from '../shared/helperFunctions.ts';
 
 export interface SubscriptionHandle {
@@ -36,6 +42,10 @@ const generateSubscriptionId = (): string => {
   return globalThis.crypto.randomUUID();
 };
 
+interface PendingRequest {
+  resolve: (value: unknown) => void;
+}
+
 export class ConnectionManager {
   private readonly socket: ClientSocketWrapper;
   private readonly stores = new Map<string, ReactiveStore>();
@@ -47,6 +57,7 @@ export class ConnectionManager {
   // its collection. Held here until the matching `ready` binds the collection,
   // then flushed into the store. See handleServerMessage.
   private pendingData: DataMsg[] = [];
+  private readonly pendingRequests = new Map<string, PendingRequest>();
 
   constructor(socket: ClientSocketWrapper) {
     this.socket = socket;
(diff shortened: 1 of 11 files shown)


KevinOrfas (changes requested):
Your PR description says the error handling is centralised, but in the code it's inline in the `POST /api/files` catch. Which did you mean? The story's Done-when 3 asks for it centralised (one place, not per route), so either way that's the gap.

What works: a `.txt` returns 400 with the EkoLiteError body and stores nothing, a `.bam` still 201s, and the new test holds all three. Moving the valid test to `.bam` was the right call.

The fix is one move. A single `server.setErrorHandler` maps `RpcError` to its status and everything else to a generic 500, and the routes just throw. `upload` already throws; the other two replies, `no file in request` (`index.ts:61`) and `file not found` (`index.ts:89`), become throws too. Then all three errors come back in the same `{ code, message }` shape, which is step c of the story done at the same time. If inline was deliberate, fair enough, but then the description shouldn't say centralised.

Two small ones, take or leave:

- Inside `if (err instanceof RpcError)` you call `toEkoLiteError(err)`, which checks the same `instanceof` again. `{ code: err.code, message: err.message }` says it directly. Keep the generic 500 though, not leaking the internal message is right.
- `'throws an ekoLiteError with a code of 400'` reads a bit off, since the route returns a 400 rather than throwing to the client.

KevinOrfas (approved):
This is it. The `setErrorHandler` is the move, the routes throw and one place renders every `EkoLiteError`, and the two old `{ error }` replies fold in. I ran the integration suite on this head: 4 passed, including the two follow-up tests now flipped to `it`, so the no-file 400 and the not-found 404 really do come back in the shared shape. Unit suite, lint and typecheck green too.

One to tidy before merge: the comment above those two tests still explains the `it.fails` handoff and tells you to flip `.fails` back to `it`. They're regular `it` now and the gap is closed, so it describes a state that no longer exists. Drop it.

One to think about, not for this PR: the handler sends `toEkoLiteError(err)` for everything, so an unexpected 500 now carries the internal `err.message` out to the client. That matches what the websocket path already does, so it's consistent rather than worse. Whether we want to sanitise 500 messages across both transports is its own conversation, a system change, not this story.

Good work, this closes Done-when 3 and step c together.

PR #88, opened 2026-06-26
Eko 205/5c1 post apiupload route

Iam-Ebuka:
## Summary

Adds a reusable `Uploader` client for file uploads backed by `XMLHttpRequest`.

The uploader exposes an `upload(file)` API that submits a file as multipart form data to `POST /api/files` and resolves with the stored file metadata (`{ id, name }`) on successful responses. It follows the existing nullable infrastructure pattern by providing `create()` for the real transport and `createNull()` for a stubbed transport used in tests.

The null implementation records outgoing upload requests and returns a configurable response, allowing the upload flow to be exercised without making network requests. The upload logic was also separated from the transport implementation to provide a seam for future upload progress and error handling.

## Tests

Added `tests/client/uploader.test.ts` to verify that:

* `upload(file)` sends a `POST` request to `/api/files`.
* The upload request is recorded by the null transport.
* A successful upload resolves with the server's `{ id, name }` response.
* The upload flow is exercised entirely through the null transport without performing network requests.

https://linear.app/ekohacks/issue/EKO-205/5c1-post-apiupload-route

### DIFF
--- a/client/uploader.ts
+++ b/client/uploader.ts
@@ -2,6 +2,12 @@ import { EventEmitter, OutputTracker } from '../server/infrastructure/outputTrac
 
 const REQUEST_EVENT = 'request';
 
+interface UploadRequest {
+  method: string;
+  url: string;
+  body: FormData;
+}
+
 interface UploadResponse {
   id: string;
   name: string;
@@ -104,22 +110,44 @@ export class Uploader {
   }
 
   async upload(file: File): Promise<UploadResponse> {
-    return new Promise((resolve, reject) => {
-      const form = new FormData();
-      form.append('file', file);
+    const request = this.buildRequest(file);
 
-      this.request.open('POST', '/api/files');
+    this.emitter.emit(REQUEST_EVENT, {
+      method: request.method,
+      url: request.url,
+    });
+
+    return this.execute(request);
+  }
+
+  private buildRequest(file: File): UploadRequest {
+    const body = new FormData();
+    body.append('file', file);
+
+    return {
+      method: 'POST',
+      url: '/api/files',
+      body,
+    };
+  }
+
+  private parseResponse(): UploadResponse {
+    const parsed: unknown = JSON.parse(this.request.responseText);
+
+    if (!isUploadResponse(parsed)) {
+      throw new Error('Invalid upload response');
+    }
+
+    return parsed;
+  }
+
+  private execute(request: UploadRequest): Promise<UploadResponse> {
+    return new Promise((resolve, reject) => {
+      this.request.open(request.method, request.url);
 
       this.request.onload = () => {
         if (this.request.status >= 200 && this.request.status < 300) {
-          const parsed: unknown = JSON.parse(this.request.responseText);
-
-          if (!isUploadResponse(parsed)) {
-            reject(new Error('Invalid upload response'));
-            return;
-          }
-
-          resolve(parsed);
+          resolve(this.parseResponse());
         } else {
           reject(new Error(`Upload failed (${String(this.request.status)})`));
         }
@@ -129,12 +157,7 @@ export class Uploader {
         reject(new Error('Upload failed'));
       };
 
-      this.request.send(form);
-
-      this.emitter.emit(REQUEST_EVENT, {
-        method: 'POST',
-        url: '/api/files',
-      });
+      this.request.send(request.body);
     });
   }
 
--- a/client/uploader.ts
+++ b/client/uploader.ts
@@ -1,0 +1,144 @@
+import { EventEmitter, OutputTracker } from '../server/infrastructure/outputTracker.ts';
+
+const REQUEST_EVENT = 'request';
+
+interface UploadResponse {
+  id: string;
+  name: string;
+}
+
+function isUploadResponse(data: unknown): data is UploadResponse {
+  if (typeof data !== 'object' || data === null) {
+    return false;
+  }
+
+  const response = data as Record<string, unknown>;
+
+  return typeof response.id === 'string' && typeof response.name === 'string';
+}
+
+interface NullUploaderOptions {
+  response: {
+    status: number;
+    body: UploadResponse;
+  };
+}
+
+interface RequestLike {
+  open(method: string, url: string): void;
+  send(body: FormData): void;
+
+  onload: (() => void) | null;
+  onerror: (() => void) | null;
+
+  status: number;
+  responseText: string;
+}
+
+class RealRequest implements RequestLike {
+  private readonly xhr = new XMLHttpRequest();
+
+  onload: (() => void) | null = null;
+  onerror: (() => void) | null = null;
+
+  constructor() {
+    this.xhr.onload = () => this.onload?.();
+    this.xhr.onerror = () => this.onerror?.();
+  }
+
+  open(method: string, url: string): void {
+    this.xhr.open(method, url);
+  }
+
+  send(body: FormData): void {
+    this.xhr.send(body);
+  }
+
+  get status(): number {
+    return this.xhr.status;
+  }
+
+  get responseText(): string {
+    return this.xhr.responseText;
+  }
+}
+
+class NullRequest implements RequestLike {
+  onload: (() => void) | null = null;
+  onerror: (() => void) | null = null;
+
+  status: number;
+  responseText: string;
+
+  method = '';
+  url = '';
+
+  constructor(private readonly response: NullUploaderOptions['response']) {
+    this.status = response.status;
+    this.responseText = JSON.stringify(response.body);
+  }
+
+  open(method: string, url: string): void {
+    this.method = method;
+    this.url = url;
+  }
+
+  send(_body: FormData): void {
+    queueMicrotask(() => {
+      this.onload?.();
+    });
+  }
+}
+
+export class Uploader {
+  private readonly emitter = new EventEmitter();
+
+  private constructor(private readonly request: RequestLike) {}
+
+  static create(): Uploader {
+    return new Uploader(new RealRequest());
+  }
+
+  static createNull(options: NullUploaderOptions): Uploader {
+    return new Uploader(new NullRequest(options.response));
+  }
+
+  async upload(file: File): Promise<UploadResponse> {
+    return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


client/uploader.ts
```diff
@@ -0,0 +1,167 @@
...
+    queueMicrotask(() => {
+      this.onload?.();
+    });
+  }
+}
+
+export class Uploader {
+  private readonly emitter = new EventEmitter();
+
+  private constructor(private readonly request: RequestLike) {}
```
KevinOrfas: Two uploads at once and the first never comes back. Failing test:

```ts
it('completes every upload when two run concurrently', async () => {
  const uploader = Uploader.createNull({
    response: { status: 201, body: { id: 'f1', name: 'sample.bam' } },
  });
  const requests = uploader.trackRequests();

  const first = uploader.upload(new File([Buffer.from('A')], 'a.bam'));
  const second = uploader.upload(new File([Buffer.from('B')], 'b.bam'));

  const [a, b] = await Promise.all([first, second]);

  expect(a).toEqual({ id: 'f1', name: 'sample.bam' });
  expect(b).toEqual({ id: 'f1', name: 'sample.bam' });
  expect(requests.data).toHaveLength(2);
}, 2000);
```

The Uploader holds one `request` for its whole life, and `execute()` reassigns `this.request.onload` (the single "response arrived" callback) on every call. So the second `upload()` overwrites the first one's callback, and the first promise sits waiting on a result that now belongs to the second. It times out.

The shape came from `ClientSocketWrapper`, which is right to hold one socket because a socket is long lived. An upload isn't: it's one request, then done. Look at how `create()` and `createNull()` take a factory in `client/clientSocket.ts:229` so each wrapper builds its own transport, and give each `upload()` its own request the same way.

One for you to decide: when each upload makes its own null request, should `createNull` replay the same response every time, or queue them the way `ConfigurableResponse` does in `server/infrastructure/outputTracker.ts:61`?

client/uploader.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  static createNull(options: NullUploaderOptions): Uploader {
+    return new Uploader(new NullRequest(options.response));
+  }
+
+  async upload(file: File): Promise<UploadResponse> {
+    const request = this.buildRequest(file);
+
+    this.emitter.emit(REQUEST_EVENT, {
```
KevinOrfas, line 117: The tracker can see a POST happened, but not which file. Failing test:

```ts
it('records which file went out, not just the route', async () => {
  const uploader = Uploader.createNull({
    response: { status: 201, body: { id: 'f1', name: 'sample.bam' } },
  });
  const requests = uploader.trackRequests();

  await uploader.upload(new File([Buffer.from('BAMDATA')], 'sample.bam'));

  expect(requests.data).toContainEqual(
    expect.objectContaining({ method: 'POST', url: '/api/files', filename: 'sample.bam' }),
  );
});
```

This emits `{ method, url }` and drops the body, so the recorded request is identical no matter what you upload. The point of an OutputTracker recording double is to assert on what left the client, the way `trackMessages()` records the actual message in `client/clientSocket.ts:331`, not just that a message went. The test already in this PR is named "sends the bytes", so this is really about making the recording match that promise.

Question: what is the smallest thing worth recording to make it useful, just the filename, or filename plus size and type?

client/uploader.ts
```diff
@@ -0,0 +1,167 @@
...
+    return parsed;
+  }
+
+  private execute(request: UploadRequest): Promise<UploadResponse> {
+    return new Promise((resolve, reject) => {
+      this.request.open(request.method, request.url);
+
+      this.request.onload = () => {
+        if (this.request.status >= 200 && this.request.status < 300) {
+          resolve(this.parseResponse());
```
KevinOrfas: A 2xx with an unexpected body hangs instead of failing. Failing test:

```ts
it('rejects, rather than hanging, when a 2xx body is the wrong shape', async () => {
  const uploader = Uploader.createNull({
    response: {
      status: 201,
      body: { wrong: 'shape' } as unknown as { id: string; name: string },
    },
  });

  await expect(
    uploader.upload(new File([Buffer.from('x')], 'x.bam')),
  ).rejects.toThrow('Invalid upload response');
}, 2000);
```

`parseResponse()` throws when the body is not `{ id, name }`, but it throws inside the `onload` callback, on a later turn of the event loop, where this Promise's `reject` cannot catch it. So the throw escapes as an unhandled error and the caller's promise never settles. That means the `isUploadResponse` guard cannot do its job yet.

The `onerror` branch just below already shows the move: it calls `reject(...)`. The success branch needs to catch the parse or validation failure and reject the same way.

Worth noting too: the null cannot even express a malformed body without that `as unknown` cast, because `response.body` is typed as the happy shape. A real server returning junk on a 2xx is something the null cannot simulate yet.

KevinOrfas (changes requested):
Nice clean seam in here, the RequestLike split reads well. Three things to iron out before this merges, each with a failing test that pins it. All three are red against this branch right now and turn green once the transport is sorted.

They cluster on one root: the uploader holds a single request and drives it through one shared `onload`, so concurrency, request recording, and the success/parse path all lean on that one seam. Comments are inline on the lines they touch.

KevinOrfas (commented):
Re-checked at `ebcf83f`. All three are genuinely sorted: the factory gives each upload its own request so concurrency holds, the tracked event carries `filename`, and the success path catches the parse failure and rejects instead of hanging. Three tests green, typecheck clean, full suite passes.

One thing left before this can merge: the lint job is red. Detail inline.

tests/client/uploader.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    expect(requests.data).toContainEqual(
+      expect.objectContaining({ method: 'POST', url: '/api/files', filename: 'sample.bam' }),
+    );
+  });
+
+  it('rejects, rather than hanging, when a 2xx body is the wrong shape', async () => {
+    const uploader = Uploader.createNull({
+      response: {
+        status: 201,
+        body: { wrong: 'shape' } as unknown as { id: string; name: string },
```
KevinOrfas: This cast is redundant now and trips lint. When you widened `body` to `unknown` (good call, that is exactly what lets the null fake a junk response) the `as unknown as { id: string; name: string }` here stopped being needed, and `@typescript-eslint/no-unnecessary-type-assertion` makes it an error, so `npm run lint` goes red. Drop it:

```ts
body: { wrong: 'shape' },
```

Full disclosure, that cast came from the test I handed over, so this one is on me. Tests stay green and lint passes once it is gone.

client/uploader.ts
KevinOrfas: You answered this one in code: each upload builds a fresh `NullRequest` from the same `options.response`, so it replays rather than queues. Good call. The why isn't on the thread yet though. Before we resolve, put it in your words here: why replay over the queue for an upload?

Iam-Ebuka: I chose to replay the configured response because each upload represents an independent HTTP request. Unlike the socket, there isn't any shared transport state between uploads, so one upload shouldn't affect another. Replaying the same response lets concurrent uploads complete independently while keeping the null deterministic. I think queueing responses makes more sense when the behaviour under test depends on successive responses, similar to ConfigurableResponse, but for this uploader story the goal is simply to replace the network with a predictable transport rather than model changing server behaviour.

client/uploader.ts
Iam-Ebuka: I think that Just adding the file name would make the recording more meaningful

tests/client/uploader.test.ts
Iam-Ebuka: thanks for explaining this one.

client/uploader.ts
Iam-Ebuka: Thanks for this, I have made he correction.

client/uploader.ts
KevinOrfas: The shape's right, but this is the principle stated back, not your read of this code, and that's the part I'm after. Make it concrete: if `createNull` queued responses instead of replaying, what would the two-upload concurrency test need to pass, and what breaks the moment a third upload runs?

client/uploader.ts
KevinOrfas: Filename's the right pick. The reasoning isn't down yet though: what does recording it let a test prove that `method` and `url` alone can't?

client/uploader.ts
KevinOrfas: Reopening this one. The fix is right, but it closed on a thanks with the why left off, and this is one worth being able to say back in your own words: why was the parse error hanging the upload instead of rejecting it?

client/uploader.ts
Iam-Ebuka: If we were going with the queue option, we would have created a configurable response instance maybe by creating a factory after that for two tests to pass we would need to add two configurable responses in createNull. If our responses are used up by the two previous uploads then the third will fail because we have used up the response.

client/uploader.ts
Iam-Ebuka: By adding the file name we can then track effectively what file left the client and not just the method and url.

client/uploader.ts
Iam-Ebuka: In the previous implementation, the parse error was hanging because the error didn't happen immediately in the promise block. It didn't happen directly because we are dealing here with asynchronous programming, we handled the reject in the onerror part but what if the error happened in onload, that where this correction comes in.

client/uploader.ts
KevinOrfas: That's the point of it. `method` and `url` are the same for every upload, so a tracker holding only those can prove a POST happened but not which file. With the filename in, the recording can tell two uploads apart, which is what lets "sends the bytes" actually assert the bytes. Resolving.

client/uploader.ts
KevinOrfas: Exactly that. The executor runs and returns, then `onload` fires on a later tick. A throw in there isn't wrapped by anything that calls `reject`, so it escapes as an unhandled error and the promise never settles. `onerror` always converted its failure into a `reject`, the success path didn't, so a parse failure had no path back to the caller. Now `onload` catches and rejects the same way. Resolving.

KevinOrfas (approved):

PR #89, opened 2026-06-29
Eko 204/5c2 route returns 400 for invalid files

Iam-Ebuka:
## Summary

This PR adds client side handling for failed file uploads by ensuring `Uploader.upload()` rejects with the server's error response instead of resolving successfully. The uploader now exposes the same `EkoLiteError` shape used by the socket client, allowing consumers to handle upload errors and RPC method errors consistently.

### Error handling

* Updated the upload flow to inspect the HTTP response status after every upload.
* Successful (`2xx`) responses continue to resolve with the parsed upload response (`{ id, name }`).
* Non successful responses now parse the server's error body and reject the promise with the shared `EkoLiteError` shape (`{ code, message }`).

### Response handling refactor

* Refactored the upload completion logic so the success or failure decision is made in a single place after reading the response status.
* Removed the previous upload specific error type in favour of the shared error contract already used by the socket client.
* This keeps error handling consistent across different client transports and provides a single way for consumers to handle failures.

### Null uploader

* Updated the null uploader so it can simulate both successful and failed server responses.
* The response body is now typed as `unknown`, allowing tests to provide either a successful upload response or an error response without relying on the network.

## Tests

Added tests covering the upload error path.

### Successful upload

Verified that:

* `Uploader.createNull()` returns the configured successful response.
* `upload()` resolves with the stored file information.
* The request tracker records the outgoing upload request.

### Failed upload

Added a test to verify that when the server returns:

```ts
{
  status: 400,
  body: {
    code: 400,
    message: "Unsupported file type: .txt",
  },
}
```

* `upload()` rejects instead of resolving.
* The rejected value matches the shared `EkoLiteError` shape.
* The error contains the server supplied `code` and `message`.

https://linear.app/ekohacks/issue/EKO-204/5c2-route-returns-400-for-invalid-files

### DIFF
--- a/client/uploader.ts
+++ b/client/uploader.ts
@@ -1,17 +1,8 @@
 import { EventEmitter, OutputTracker } from '../server/infrastructure/outputTracker.ts';
 import { EkoLiteError } from '../shared/protocol.ts';
+import { RpcError } from '../shared/types.ts';
 
 const REQUEST_EVENT = 'request';
-
-class UploadErrorResponse extends Error implements EkoLiteError {
-  constructor(
-    public readonly code: number,
-    message: string,
-  ) {
-    super(message);
-    this.name = 'UploadError';
-  }
-}
 
 function isUploadError(data: unknown): data is EkoLiteError {
   if (typeof data !== 'object' || data === null) {
@@ -170,7 +161,7 @@ export class Uploader {
       throw new Error('Invalid upload error');
     }
 
-    throw new UploadErrorResponse(parsed.code, parsed.message);
+    throw new RpcError(parsed.code, parsed.message);
   }
 
   private execute(upload: UploadRequest): Promise<UploadResponse> {
--- a/client/uploader.ts
+++ b/client/uploader.ts
@@ -2,6 +2,12 @@ import { EventEmitter, OutputTracker } from '../server/infrastructure/outputTrac
 
 const REQUEST_EVENT = 'request';
 
+interface UploadRequest {
+  method: string;
+  url: string;
+  body: FormData;
+}
+
 interface UploadResponse {
   id: string;
   name: string;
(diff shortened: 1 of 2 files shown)


client/uploader.ts
```diff
@@ -0,0 +1,201 @@
+import { EventEmitter, OutputTracker } from '../server/infrastructure/outputTracker.ts';
+import { EkoLiteError } from '../shared/protocol.ts';
+
+const REQUEST_EVENT = 'request';
+
+class UploadErrorResponse extends Error implements EkoLiteError {
```
KevinOrfas: This class is a copy of `RpcError` in `shared/types.ts:73`. Same `extends Error implements EkoLiteError`, same `code` and `message`, only the `name` is different.

Here is the bit worth seeing: the 400 you catch here was born as an `RpcError` on the server. `fileUploadError()` in `shared/types.ts:99` builds `new RpcError(400, 'Unsupported file type...')`, the error handler flattens it to `{ code, message }` on the wire, and over here you build a fresh `UploadErrorResponse` that means the same thing. So one error type goes out, gets lost in transit, and a second one is created to replace it.

If the client rehydrates it back into `RpcError`, the duplicate class goes and there is one error type across the whole app. A test that pins it:

```ts
import { RpcError } from '../../shared/types.ts';

it('rejects with the shared RpcError the rest of the app already uses', async () => {
  const uploader = Uploader.createNull({
    response: { status: 400, body: { code: 400, message: 'Unsupported file type: .txt' } },
  });

  await expect(
    uploader.upload(new File([Buffer.from('notes')], 'bad.txt')),
  ).rejects.toBeInstanceOf(RpcError);
});
```

Red right now (the rejection is an `UploadError`), green once `resolveUpload` throws `new RpcError(parsed.code, parsed.message)` and the local class is gone. I ran it: the existing error path test stays green and typecheck is clean.

One real question, separate from the cleanup: the socket client does not do this either. It rejects RPC errors with the bare `message.error` object (`connectionManager.ts:286`), not an `Error` at all. So "consistent with the socket client" is not quite true yet. Do you want to bring the socket path onto `RpcError` too, or is that its own story?

KevinOrfas (changes requested):
One thing to tidy before merge: the upload error path mints a second error type for something the app already has. Backed by a failing test inline. Leaving the lint cast out of this review since you are sorting it in #88.

client/uploader.ts
Iam-Ebuka: oh noted. I see it

KevinOrfas: Reopening. The cleanup landed, `RpcError` rehydrated, the local class gone, the test pinning it green, so the code's right. Two things kept it from being properly resolved though.

First, the why never made it onto the thread. Spotting the duplicate and understanding why it's worth removing aren't the same thing, and only the second one closes this. In your own words: why is minting a fresh error on the client a problem when the server already sent one as an `RpcError`?

Second, and separate: you didn't answer the question I asked. The socket client still rejects with a bare `message.error` object at `connectionManager.ts:286`, not an `RpcError`, so the app isn't consistent yet. That's a real call, bring the socket path onto `RpcError`, or keep it as its own story. If it's its own story it needs a Linear issue so it doesn't vanish. Either way I need the decision, not a resolve.

Iam-Ebuka: Alright, so for the first one what happens is the the server already returns an instance for RpcError so when it returns to the client, instead of inventing a new class, why not just use what we already have. So the new test that was added failed with the previous implementation because it wasn't directly an instance of RpcError.

For the second, I honestly missed this question the first time, sorry about that. I would suggest we make it its own story.

KevinOrfas: That's it, and the decision's the right shape too. One refinement so the model's exact: the `RpcError` instance doesn't actually reach the client. It's flattened to a plain `{ code, message }` on the wire, the class identity is lost in transit, and `resolveUpload` rebuilds a new `RpcError` from that data. So it's less "use the instance the server sent" and more "the server's error arrives as plain data and we reconstruct it into the one error type the app shares", which is what makes the `instanceof` check pass again. The reuse instinct was bang on, that's just the mechanism underneath it.

On the socket path, agreed, its own story. Raised it as EKO-287 so it doesn't slip. Resolving this. No need to reply, just wanted the wire bit on record.

KevinOrfas: https://linear.app/ekohacks/issue/EKO-287/socket-client-should-reject-rpc-errors-as-rpcerror-not-bare-wire

KevinOrfas (approved):

PR #93, opened 2026-07-01
Eko 202/5d1 report upload progress

Iam-Ebuka:
## Summary

This PR adds upload progress reporting to the client uploader while keeping the upload transport reusable.

The uploader now exposes an optional `onProgress` callback that reports upload progress as a normalized `{ percent }` value. The real implementation is wired to `XMLHttpRequest.upload.onprogress`, while the null uploader can replay configured progress events so the behaviour can be tested without making a real network request.

## What changed

* Added upload progress support to the uploader API via an optional `onProgress` callback.
* Extended `RequestLike` with a progress handler so both the real and null implementations expose the same interface.
* Wired `RealRequest` to `XMLHttpRequest.upload.onprogress`.
* Updated `NullRequest` to replay configured progress events before completing the upload, allowing progress behaviour to be tested without a browser.
* Moved progress normalization into a single helper that converts the raw `loaded` and `total` values into a `{ percent }` object.
* Added a guard for uploads where the total size is unknown (`lengthComputable === false`) to avoid dividing by zero and to keep progress reporting consistent.

## Tests

Added tests covering the new upload progress behaviour:

* Upload progress callbacks are fired as progress events are received.
* Progress values are reported as normalized percentages.
* The null uploader replays configured progress events before resolving the upload.
* Uploads whose total size is unknown (`lengthComputable === false`) do not attempt to calculate a percentage and still complete successfully.

https://linear.app/ekohacks/issue/EKO-202/5d1-report-upload-progress

### DIFF
--- a/client/uploader.ts
+++ b/client/uploader.ts
@@ -55,11 +55,7 @@ interface NullUploaderOptions {
     body: unknown;
   };
 
-  progress?: {
-    loaded: number;
-    total: number;
-    lengthComputable?: boolean;
-  }[];
+  progress?: ProgressEventLike[];
 }
 
 interface RequestLike {
@@ -126,7 +122,7 @@ class NullRequest implements RequestLike {
 
   constructor(
     response: NullUploaderOptions['response'],
-    private readonly progress: NullUploaderOptions['progress'] = [],
+    private readonly progress: ProgressEventLike[] = [],
   ) {
     this.status = response.status;
     this.responseText = JSON.stringify(response.body);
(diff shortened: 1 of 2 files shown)


client/uploader.ts
```diff
@@ -40,6 +54,8 @@ interface NullUploaderOptions {
     status: number;
     body: unknown;
   };
+
+  progress?: ProgressEventLike[];
```
KevinOrfas, line 58: This was an inline object type identical to `ProgressEventLike` above, so I pointed it at `ProgressEventLike[]` — one shape to maintain instead of two that can drift.

client/uploader.ts
```diff
@@ -105,6 +135,14 @@ class NullRequest implements RequestLike {
 
   send(_body: FormData): void {
     queueMicrotask(() => {
+      for (const step of this.progress) {
```
KevinOrfas, line 140: `progress` defaults to `[]` in the constructor, so the old `if (this.progress)` guard was always true and looping an empty array is already a no-op. Dropped it and typed the field as a definite `ProgressEventLike[]` so it stays non-nullable. Typecheck and the uploader tests stay green.

KevinOrfas (approved):

client/uploader.ts
Iam-Ebuka: Oh that's true, everything in there is the same as ProgressEvenLike so no need for duplicates.

PR #116, opened 2026-07-06
Eko 292/7b4 refactor shutdown and processwrapper

Iam-Ebuka:
**Before the review below, here are the answers to the three questions ask:**

1. If StubbedProcess.exit() ever throws an error, "it records exits in an output tracker instead of killing the process" will fail because a throw when in an outputTracker terminates the tracker process.
2. trackExits() works on both real and null instances because it is used for tracking our events. It doesn't mutate any process behavior but  advanceTime() works only on stubbed because we use it to force time to move forward in test and simulateSignal() injects a signal into the system. Time cannot be manually advanced in real Node.js and also the signal will come from the operating system and not the test code.
3.  The rule we follow that decides which folder a class belongs in is the A-frame by James Shore that imposes that Application, Infrastructure and logic should be separated for nullable testing without mocks. 

### Overview

This refactor improves the process abstraction and shutdown flow by removing duplicated logic, tightening stub correctness, and improving internal consistency between real and stubbed implementations. No production behavior changes were introduced except where explicitly noted as previously undefined stub behavior now being pinned by tests.

### What changed
**1. Duplicated guard removed in ProcessWrapper**
- Extracted shared instanceof StubbedProcess check into requireStubbedProcess()
- Used in advanceTime() and simulateSignal()
- Preserves exact existing error messages
- Improves consistency and reduces duplication

**2. Shutdown signal handler extracted**
- Moved inline signal callback logic into handleShutdownSignal()
- arm() now only registers the listener
- No behavioral changes

**3. Shutdown logging decision**
- Kept both console.error calls inside shutdown flow
- Chose not to introduce a logging seam
- Reason: logs are operational signals (timeouts/failures), not test-relevant output
- Accepts minor test output noise for simplicity and production observability

**4. StubbedProcess timer correctness fix**
- Timers now execute in deadline order instead of insertion order
- Fired/cancelled timers are properly cleaned up after execution
- Prevents internal timer growth and improves stub fidelity
- New test added to lock in ordering behavior (no existing test modified)

**5. StubbedProcess.exit consistency**
Aligned exit(code: number) signature with ProcessLike and RealProcess
Stub still ignores the code value intentionally
Added comment clarifying omission is intentional for test simulation

### Tests
- All existing tests pass without modification
- One new test added to verify timer execution ordering
- No behavior changes introduced to existing test contracts

### Summary
- Reduced duplication in process abstraction
- Improved clarity in shutdown flow
- Tightened stub correctness (timers + API consistency)
- Preserved production behavior and observability
- Added minimal test coverage for previously unspecified stub behavior

https://linear.app/ekohacks/issue/EKO-292/7b4-refactor-shutdown-and-processwrapper

### DIFF
--- a/server/infrastructure/process.ts
+++ b/server/infrastructure/process.ts
@@ -26,6 +26,14 @@ export class ProcessWrapper {
     return new ProcessWrapper(new StubbedProcess());
   }
 
+  private requireStubbedProcess(errorMethod: 'advanceTime' | 'simulateSignal'): StubbedProcess {
+    if (!(this.proc instanceof StubbedProcess)) {
+      throw new Error(`${errorMethod} only available on null instance`);
+    }
+
+    return this.proc;
+  }
+
   onSignal(handler: (signal: Signal) => void): void {
     this.proc.onSignal(handler);
   }
@@ -44,17 +52,11 @@ export class ProcessWrapper {
   }
 
   advanceTime(ms: number): void {
-    if (!(this.proc instanceof StubbedProcess)) {
-      throw new Error('advanceTime only available on null instance');
-    }
-    this.proc.advanceTime(ms);
+    this.requireStubbedProcess('advanceTime').advanceTime(ms);
   }
 
   simulateSignal(signal: Signal): void {
-    if (!(this.proc instanceof StubbedProcess)) {
-      throw new Error('simulateSignal only available on null instance');
-    }
-    this.proc.simulateSignal(signal);
+    this.requireStubbedProcess('simulateSignal').simulateSignal(signal);
   }
 }
 
--- a/server/infrastructure/process.ts
+++ b/server/infrastructure/process.ts
@@ -109,22 +109,34 @@ class StubbedProcess implements ProcessLike {
   }
 
   advanceTime(ms: number): void {
-    this.now += ms;
+    const target = this.now + ms;
 
-    const dueTimers = this.timers
-      .filter((timer) => timer.live && timer.dueAt <= this.now)
-      .sort((left, right) => left.dueAt - right.dueAt);
+    while (true) {
+      const nextTimer = this.timers.reduce<StubbedTimer | null>((earliest, timer) => {
+        if (!timer.live || timer.dueAt > target) {
+          return earliest;
+        }
 
-    for (const timer of dueTimers) {
-      if (!timer.live) {
-        continue;
+        if (earliest === null || timer.dueAt < earliest.dueAt) {
+          return timer;
+        }
+
+        return earliest;
+      }, null);
+
+      if (nextTimer === null) {
+        this.now = target;
+        break;
       }
 
-      timer.live = false;
-      timer.callback();
-    }
+      this.now = nextTimer.dueAt;
+      nextTimer.live = false;
 
-    this.timers.splice(0, this.timers.length, ...this.timers.filter((timer) => timer.live));
+      // Remove fired/cancelled timers before running callbacks.
+      this.timers.splice(0, this.timers.length, ...this.timers.filter((timer) => timer.live));
+
+      nextTimer.callback();
+    }
   }
 
   simulateSignal(signal: Signal): void {
--- a/server/infrastructure/process.ts
+++ b/server/infrastructure/process.ts
@@ -96,8 +96,8 @@ class StubbedProcess implements ProcessLike {
     this.handlers.push(handler);
   }
 
-  exit(): void {
-    // The nulled process survives its own exit; the tracker holds the story.
+  exit(_code: number): void {
+    // Intentionally ignored. The stubbed process survives its own exit; the tracker holds the story.
   }
 
   startTimer(ms: number, callback: () => void): () => void {
--- a/server/infrastructure/process.ts
+++ b/server/infrastructure/process.ts
@@ -130,10 +130,9 @@ class StubbedProcess implements ProcessLike {
       }
 
       this.now = nextTimer.dueAt;
-      nextTimer.live = false;
 
-      // Remove fired/cancelled timers before running callbacks.
-      this.timers.splice(0, this.timers.length, ...this.timers.filter((timer) => timer.live));
+      const index = this.timers.indexOf(nextTimer);
+      this.timers.splice(index, 1);
 
       nextTimer.callback();
     }
--- a/server/infrastructure/process.ts
+++ b/server/infrastructure/process.ts
@@ -26,9 +26,9 @@ export class ProcessWrapper {
     return new ProcessWrapper(new StubbedProcess());
   }
 
-  private requireStubbedProcess(errorMethod: 'advanceTime' | 'simulateSignal'): StubbedProcess {
+  private requireStubbedProcess(): StubbedProcess {
     if (!(this.proc instanceof StubbedProcess)) {
-      throw new Error(`${errorMethod} only available on null instance`);
+      throw new Error('Method only available on null instance');
     }
 
     return this.proc;
@@ -52,11 +52,11 @@ export class ProcessWrapper {
   }
 
   advanceTime(ms: number): void {
-    this.requireStubbedProcess('advanceTime').advanceTime(ms);
+    this.requireStubbedProcess().advanceTime(ms);
   }
 
   simulateSignal(signal: Signal): void {
-    this.requireStubbedProcess('simulateSignal').simulateSignal(signal);
+    this.requireStubbedProcess().simulateSignal(signal);
   }
 }
 
--- a/server/infrastructure/process.ts
+++ b/server/infrastructure/process.ts
@@ -110,12 +110,21 @@ class StubbedProcess implements ProcessLike {
 
   advanceTime(ms: number): void {
     this.now += ms;
-    for (const timer of [...this.timers]) {
-      if (timer.live && timer.dueAt <= this.now) {
-        timer.live = false;
-        timer.callback();
+
+    const dueTimers = this.timers
+      .filter((timer) => timer.live && timer.dueAt <= this.now)
+      .sort((left, right) => left.dueAt - right.dueAt);
+
+    for (const timer of dueTimers) {
+      if (!timer.live) {
+        continue;
       }
+
+      timer.live = false;
+      timer.callback();
     }
+
+    this.timers.splice(0, this.timers.length, ...this.timers.filter((timer) => timer.live));
   }
 
   simulateSignal(signal: Signal): void {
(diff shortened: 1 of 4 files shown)


server/infrastructure/process.ts
```diff
@@ -26,6 +26,14 @@ export class ProcessWrapper {
     return new ProcessWrapper(new StubbedProcess());
   }
 
+  private requireStubbedProcess(errorMethod: 'advanceTime' | 'simulateSignal'): StubbedProcess {
```
KevinOrfas: The extraction removes the duplicated `instanceof` check, which is good. But the string parameter creates a new problem: each caller has to pass its own method name, and nothing checks it matches. `requireStubbedProcess('advanceTime')` inside `simulateSignal` compiles fine and produces an error that blames the wrong method. No test pins these messages (process.test.ts has no `toThrow`), so they're free to change.

Have a look at how the other wrappers solve the same problem: `websocket.ts:88` and `clientSocket.ts:337` both do the check inline with a fixed message. With only two call sites, inline is the same amount of code and there's no union type to maintain. If you prefer to keep the helper, use one fixed message like `'only available on null instance'` so the caller can't pass the wrong name. No test to write here; the existing suite staying green is the proof.

server/infrastructure/process.ts
```diff
@@ -108,12 +110,21 @@ class StubbedProcess implements ProcessLike {
 
   advanceTime(ms: number): void {
     this.now += ms;
-    for (const timer of [...this.timers]) {
-      if (timer.live && timer.dueAt <= this.now) {
-        timer.live = false;
-        timer.callback();
+
+    const dueTimers = this.timers
```
KevinOrfas: The deadline ordering fix is right, and your new test pins it. There's one case the rewrite keeps from the old loop, and it's worth fixing while you're here: a timer created inside another timer's callback never fires in the same `advanceTime` call. The real process would fire it. With `setTimeout`, a 5ms timer created inside a 10ms callback fires at 15ms.

Failing test for process.test.ts:

```ts
it('fires a timer scheduled by another timer, when both come due in the same advance', () => {
  const proc = ProcessWrapper.createNull();
  const order: number[] = [];

  proc.startTimer(10, () => {
    order.push(10);
    proc.startTimer(5, () => order.push(15));
  });

  proc.advanceTime(100);

  expect(order).toEqual([10, 15]);
});
```

Why it fails: `this.now += ms` runs first, so `now` is already 100 when the outer callback runs. The nested timer gets `dueAt: 105` and misses. Checking the array again after the loop won't help, because 105 is still in the future.

Hint: the stub needs to move time step by step instead of jumping to the end. Save the target first (`const target = this.now + ms`). Then loop: find the earliest live timer with `dueAt <= target`, set `this.now` to that timer's `dueAt`, fire it. When no timer qualifies, set `this.now = target` and stop. With that loop the sort and the `!timer.live` recheck are no longer needed. This is how vitest's fake timers work.

server/infrastructure/process.ts
```diff
@@ -108,12 +110,21 @@ class StubbedProcess implements ProcessLike {
...
+    for (const timer of dueTimers) {
+      if (!timer.live) {
+        continue;
       }
+
+      timer.live = false;
+      timer.callback();
     }
+
+    this.timers.splice(0, this.timers.length, ...this.timers.filter((timer) => timer.live));
```
KevinOrfas: This line has no observable effect. `timers` is private, and the only code that reads it is the filter above, which already skips dead timers. The old code never pruned and every test passed the same way. If no test can fail because of a line, the refactor phase says the line goes.

If you do want to keep the array small, which is a fair instinct, the simpler form is to drop `readonly` from the field and reassign: `this.timers = this.timers.filter((timer) => timer.live);`. The splice with a spread of a filter of the same array is hard to read and does the same thing. One more detail worth knowing: `cancel()` only sets `live` to false, so a cancelled timer stays in the array until the next `advanceTime`. If you build the loop from the other thread, fired timers can be removed as they fire and this line disappears on its own.

server/logic/shutdown.ts
```diff
@@ -23,27 +23,31 @@ export class Shutdown {
...
+    }
+    this.shuttingDown = true;
+    const cancelDeadline = this.proc.startTimer(this.graceMs, () => {
+      console.error('shutdown timed out, exiting hard');
+      this.proc.exit(1);
     });
+    void this.closable.close().then(
+      () => {
+        cancelDeadline();
+        this.proc.exit(0);
```
KevinOrfas, line 58: This bug is older than your PR. The move just made it visible, so let's fix it here. The problem: when the deadline fires first, `exit(1)` runs, but the pending `close()` promise is still out there. When it eventually resolves, this handler runs anyway. `cancelDeadline()` does nothing (the timer already fired) and then `exit(0)` records a clean exit after a timed out one. In production the process is already dead so nobody sees it. On the nulled process, which survives its own exit by design (process.ts:100), the tracker ends up with `[{ code: 1 }, { code: 0 }]`. That says the shutdown timed out and then also went cleanly, which can't both be true.

Failing test for shutdown.test.ts, using the existing helpers:

```ts
it('a close that resolves after the deadline does not record a second, clean exit', async () => {
  const { closable, resolveClose } = closableDouble();
  const proc = ProcessWrapper.createNull();
  const exits = proc.trackExits();
  new Shutdown(closable, proc).arm();

  proc.simulateSignal('SIGTERM');
  proc.advanceTime(5000);
  expect(exits.data).toEqual([{ code: 1 }]);

  resolveClose();
  await flush();

  expect(exits.data).toEqual([{ code: 1 }]);
});
```

Hint: `shuttingDown` can't guard this. It's already true on this path, and it has a different job (the second signal branch). The class needs one more boolean that means an exit already happened. Set it where the deadline calls `exit(1)`, check it at the top of both close handlers.

KevinOrfas (commented):
Solid refactor. I traced every line you deleted to where its behaviour lands in the new code and nothing was lost: the guard errors are identical, the deadline ordering now matches the real process, and your new test pins it.

Four threads inline, all to be fixed in this PR. Two are simplifications of new code (the guard helper's string parameter, the pruning splice); the existing suite staying green is the proof for those. Two are real gaps with a failing test each that you can paste in: a timer created by another timer never fires in the same advance, and a shutdown that timed out records a second, clean exit. Make those tests pass and the claims are settled. As you close each thread, write a line about what you found so we resolve on understanding, not just on green tests.

server/infrastructure/process.ts
Iam-Ebuka: I think going back to how the other wrappers solved the problem is us not needing the refactor solution any more because they check inline which is what we did before the refactor. The best way to solve this problem is to remove the parameter to that it will just throw a fixed error message..

server/infrastructure/process.ts
Iam-Ebuka: Okay, from what I understand from this problem pointed out, here is the explanation of what is currently happening, and then the pattern in which i would follow to solve the problem. Currently if proc.advanceTime = 100 and now = 0, when the stub runs the method it instantly makes now 100 without following any process or a count because of the first line in the method ` this.now += ms` . And when it goes to the callback and it sees the next timer at 5, but now is already 100 and therefore the loop ends.

 To solve this problem, I would be following the pseudo code as stated below:  

```
target = now + ms

while (true)

    find earliest live timer
    whose dueAt <= target

    if none

        now = target
        break

    now = timer.dueAt

    remove timer

    callback()
```

server/infrastructure/process.ts
Iam-Ebuka: I guess you are referring to the splice that was in line 127, it has no effect on if our test passes or not because timer.live are checked and already ignored earlier on.

In my previous commit, I changed the line to the below which solved that problem
```
     nextTimer.live = false;

      // Remove fired/cancelled timers before running callbacks.
      this.timers.splice(0, this.timers.length, ...this.timers.filter((timer) => timer.live));

```

In other word, it removes the cancelled timers after the update but I will futher refactor it by removing it from the array instantly instead of having to reassign it to false before removing. The below code will be the shape: 


```
const index = this.timers.indexOf(nextTimer);
this.timers.splice(index, 1);

nextTimer.callback();

```

now the fired timer is gone immediately.

server/logic/shutdown.ts
Iam-Ebuka: Yes I get what you mean by shuttingDown can't guard this because shuttingDown askes  the question "have we started shutting down?" not "have we already exited?" . I will introduce a Boolean as suggested to ask "has an exit already happened?" so that the it would guard to prevent the second from firing when it resolves.

KevinOrfas: The flag fixes the case and the test is green. There's a second path it doesn't cover yet.

The second signal branch (the early `exit(1)` when `shuttingDown` is already true) also exits, but it never sets `exited`. So this sequence still records two exits:

```ts
it('a second signal exits hard, and a later clean close records no second exit', async () => {
  const { closable, resolveClose } = closableDouble();
  const proc = ProcessWrapper.createNull();
  const exits = proc.trackExits();
  new Shutdown(closable, proc).arm();

  proc.simulateSignal('SIGTERM');
  proc.simulateSignal('SIGTERM');
  expect(exits.data).toEqual([{ code: 1 }]);

  resolveClose();
  await flush();

  expect(exits.data).toEqual([{ code: 1 }]);
});
```

Paste it in and it fails with `[{ code: 1 }, { code: 0 }]`: the second signal exits 1, then the original close resolves and the success handler records exit 0 on top. Same problem as the deadline case, different trigger.

The cause is that the flag guards two of the exits, not all of them. This class has four exits (second signal, deadline, close resolves, close rejects) and only two check the flag. Rather than add the check in two more places, put every exit behind one gate:

```ts
private exitOnce(code: number): void {
  if (this.exited) {
    return;
  }
  this.exited = true;
  this.proc.exit(code);
}
```

Then the second signal branch, the deadline, and both close handlers call `this.exitOnce(...)`, and the "already exited" check lives in one place instead of at each site. I ran the full shutdown suite plus the test above with that change: all fifteen pass.

Fixing it here. When you push, say back what the second signal path was doing before the gate. That's what resolves the thread, not the green.

Iam-Ebuka: Before the gate, the second signal what calling exit() directly without checking if it had already exited or not, it didn't also make sure to mark the exited as true. This resulted to exit(0) being fired after in resolved.

KevinOrfas (approved):
Approving. Everything traces. The guard helper and the extracted signal handler are behaviour for behaviour the same as before, and the two real bugs this refactor turned up, the nested timer that never fired and the double exit after a hard one, are both fixed and pinned with tests. Clean run through red then green. Nice work.

server/logic/shutdown.ts
KevinOrfas: That's it exactly. Before the gate the second signal exited without marking `exited`, so the still pending close resolved a clean exit(0) on top of the hard one. Every exit path is guarded now and the second signal test pins it. Resolved, thanks.

PR #124, opened 2026-07-09
Eko 298/7b5 refactor closeall to asyncdisposablestack

Iam-Ebuka:
**Answer to question asked before the implementation.**
1. In the App, there are resources that rely on another basically higher-level components depending on lower-level resources, so they arranged in such a way that are shut down before the resources they rely on. AsyncDisposeableStack uses LIFO because resources are typically disposed in the reverse order they were acquired.
2. There is a difference between the runtime (Node.js) and the TypeScript compiler. TypeScript is not aware of our platform APIs so to make it aware, we use the lib. "ESNext.Disposable" in the lib adds the type definitions for AsnycDisposableStack.
3. closeAll always rejected with the first error it encountered, whereas AsyncDisposableStack.disposeAsync() rejects with a SuppressedError when multiple disposers fail, preserving both the most recent error and the earlier suppressed one.

### Overview

This PR replaces the custom closeAll helper used during application shutdown with the built in AsyncDisposableStack, updates the project to require Node.js 24, and moves the shutdown behavior tests from the helper to App.close() before removing the helper.

### What changed

- Added a new test in tests/server/app.test.ts to pin the shutdown behavior on App.close(), ensuring all cleanup steps still run even if an earlier one fails.
- Replaced `closeAll` in `App.close()` with `AsyncDisposableStack`
- App.close() now throws a SuppressedError when cleanup steps fail, whereas it previously rethrew the first error.
- Removed closeAll and its associated tests once the behavior was pinned at the application level.
- Updated CI to run on Node 24 and raised the minimum supported Node version in package.json to >=24, intentionally dropping Node 22 support in order to use the standard `AsyncDisposableStack` API.
- Added "ESNext.Disposable" to the TypeScript lib configuration while preserving the existing standard libraries. This provides type information only and does not polyfill or change the emitted JavaScript.
- Exported `flattenSuppressed` from the package entry alongside `App`, so a consumer who catches what `App.close()` rejects with can actually read it.

### Breaking change for consumers

`App` is exported from `server/index.ts`, which is the main entry of the published package, so `App.close()` is public API and what it throws has changed here. Anyone with a `catch` around it will see a `SuppressedError` where they used to see the first failure.

A `SuppressedError` keeps its causes on `.error` and `.suppressed`, and the spec makes both non-enumerable. Printing one shows a message and a stack and nothing else, so the failures underneath are invisible to `console.error`, `Object.keys` and `JSON.stringify`. `flattenSuppressed` is the walk that reads them, which is why it now leaves the package through the same door `App` does:

```ts
import { App, flattenSuppressed } from 'ekolite';

try {
  await app.close();
} catch (err) {
  console.error('close failed:', flattenSuppressed(err));
}
```

### The Node floor is documented, not enforced

`engines` now says `>=24`, and `.npmrc` sets `engine-strict=true`. That makes `npm install` fail inside this repo on an older Node, which protects us, but it does not travel with the published package: npm reads `engine-strict` from the config of whoever is running the install, and it never ships a dependency's `.npmrc`.

So someone installing EkoLite on Node 22 gets a warning they will probably skim past, then a clean install, a server that boots fine, and `AsyncDisposableStack is not defined` at shutdown. Node 22 is in maintenance until April 2027, so this is a real trade and we are making it on purpose: the standard API is worth the floor.

https://linear.app/ekohacks/issue/EKO-298/7b5-refactor-closeall-to-asyncdisposablestack


server/app.ts
```diff
@@ -7,7 +7,6 @@ import { RpcHandler } from './logic/rpcHandler.ts';
 import { Methods } from './logic/methods.ts';
 import { Files } from './logic/files.ts';
 import { defineRunCountC } from './logic/analysis.ts';
-import { closeAll } from '../shared/helperFunctions.ts';
```
KevinOrfas, line 10: The import is gone and `tests/shared/closeAll.test.ts` is gone. The function is not.

```
$ git grep -n closeAll
shared/helperFunctions.ts:9:export async function closeAll(steps: (() => Promise<void>)[]): Promise<void> {
```

One line, the definition, no callers. The description says it was removed, so I read this as a step that got away rather than a decision.

It matters more than a tidy up. Deleting the test file is safe **because** the function goes with it. What we have instead is an exported helper in a shared module that nothing calls and nothing covers, and its doc comment still describes how `App.close` behaves, which is no longer true. The next person who needs a best effort close will find it, read that comment, and believe it.

Delete `closeAll` from `shared/helperFunctions.ts`.

server/app.ts
```diff
@@ -101,13 +103,14 @@ export class App {
...
-    // close never leaves the Mongo connection open. closeAll rethrows the first error
-    // so a hung or broken shutdown still exits non-zero through Shutdown.
-    await closeAll([
-      () => this.ws.close(),
-      () => this.publications.stopAll(),
-      () => this.mongo.close(),
-    ]);
+    const stack = new AsyncDisposableStack();
+    // Register the disposals in reverse order of what we want to happen
+    //because AsyncDisposableStack operated in LIFO order.
```
KevinOrfas: `//because` is missing its space and `operated` wants to be `operates`. `npm run format:check` passed because prettier reformats code and leaves the inside of comments alone, so this class of thing only ever gets caught by a reader.

Three lines are also saying one thing. Something like:

```ts
// Disposal runs in reverse registration order, so register the teardown backwards:
// ws closes first, mongo last.
```

server/app.ts
```diff
@@ -101,13 +103,14 @@ export class App {
...
+    // The last registered is the first to be disposed.
+    stack.defer(() => this.mongo.close());
+    stack.defer(() => this.publications.stopAll());
+    stack.defer(() => this.ws.close());
+
+    await stack.disposeAsync();
```
KevinOrfas, line 117: Question 3 asked two things: what `SuppressedError` is, and what `Shutdown` then does with it. The description answers the first. `disposeAsync()` does reject with a `SuppressedError` that holds both failures. The second half is where it turns.

`server/logic/shutdown.ts` does this with a rejected close:

```ts
console.error('shutdown failed:', err);
this.proc.exit(1);
```

`console.error` prints an error's message, its stack, and its **enumerable** own properties. The spec defines `SuppressedError`'s `.error` and `.suppressed` as non enumerable. So on Node 24 an operator gets exactly this, and nothing else:

```
shutdown failed: SuppressedError: An error was suppressed during disposal
    at AsyncDisposableStack.disposeAsync (<anonymous>)
```

`socket close failed` and `mongo close failed` are both sitting on that object. Neither reaches the log. `Object.keys(err)` is `[]`.

The exit code is still 1, so the process contract has not moved. What has moved is that on the one path where the server is already having a bad night, we can no longer say why. `closeAll` rethrew the first error and the log named it.

So "behaviour stays identical" holds when one closer fails, and does not hold when two do, which is the case this whole construct exists for.

Two ways out that I can see, and the choice is a real one:

1. Keep first error semantics inside `App.close`, catch the `SuppressedError` and rethrow the root of the chain.
2. Let it through and teach `Shutdown` to walk `.error` and `.suppressed` when it logs.

I lean towards 2. It is the smaller change, and it keeps `App.close` honest about the fact that two things broke rather than quietly picking one. But you have been living in this code for a week and I have not, so tell me which you would rather defend.

Either way it wants the pin in the `tests/server/app.test.ts` thread. Shutdown internals were out of scope for the story because 7.B.4 was in flight, and that has since merged, so this is ours to finish here rather than leave for later.

tests/server/app.test.ts
```diff
@@ -47,6 +48,17 @@ describe('App.createNull assembles all parts', () => {
     expect(await app.methods.call('echo', ['ping'])).toBe('echo: ping');
   });
 
+  it('still closes Mongo when the nulled socket close rejects', async () => {
```
KevinOrfas, line 52: This is the best thing in the diff. You needed a seam that did not exist, you grew one instead of reaching for a spy, and you drove the rejecting close through `ConfigurableResponse`, which is the same mechanism `MongoWrapper.createNull({ find })` has used all along. That is the Nullable pattern working as intended.

There is a gap underneath it. `tests/shared/closeAll.test.ts` pinned two behaviours, not one:

1. every step runs even when an earlier one rejects
2. the first error is rethrown, so the caller can say what broke

This test carries (1) across. Nothing carries (2), and (2) is the one that changed.

I ran your test against the old `closeAll` and it passes there too. That is not a criticism, the story called it a pin rather than a red test and a pin is supposed to pass both sides. It does mean it cannot tell the two implementations apart, so it cannot notice what the swap did.

Here is one that can. Same test, same seams, run twice:

```ts
it('rejects with the first failure, so Shutdown can name it', async () => {
  const mongo = MongoWrapper.createNull({ close: [new Error('mongo close failed')] });
  const ws = WebSocketWrapper.createNull({ close: [new Error('socket close failed')] });
  const app = App.createNull({ mongo, ws });

  await expect(app.close()).rejects.toThrow('socket close failed');
});
```

```
with closeAll               ✓ passes
with AsyncDisposableStack   × expected 'socket close failed'
                              received 'An error was suppressed during disposal'
```

Passing on `main` and failing here is what makes it evidence rather than an opinion. What the failure means is in the thread on `server/app.ts`.

It needs one seam that does not exist yet: `MongoWrapper.createNull({ close })`. `MongoWrapper`'s private constructor already takes a `closer`, so it is a `ConfigurableResponse` and about four lines, the mirror of what you just built in `NullConnectionSource`.

If you land on option 2 in the other thread, this test says something different and probably lives next to the `Shutdown` tests instead. Write the one that pins the answer you choose.

server/infrastructure/websocket.ts
```diff
@@ -150,6 +152,10 @@ export class WebSocketWrapper implements WebSocketInterface {
     return new OutputTracker(this.emitter, MESSAGE_EVENT);
   }
 
+  trackClose(): OutputTracker {
```
KevinOrfas: `MongoWrapper.trackClose()` earns its place, the new pin uses it. This one has no caller anywhere in the repo. `noUnusedLocals` does not look at public methods, so nothing will ever tell us it has rotted.

There are four seams in play here, an input and an output on each wrapper. You added three:

| | input, make close fail | output, observe close |
|---|---|---|
| `WebSocketWrapper` | `createNull({ close })` ✓ | `trackClose()` ✓ but unused |
| `MongoWrapper` | missing | `trackClose()` ✓ |

The one you did not add is the one the second pin needs. Swap the unused one for the missing one and the diff is the same size with nothing dead in it.

tsconfig.json
```diff
@@ -31,5 +31,6 @@
     "learning/**/*.ts",
     "docs/api/__examples__/**/*.ts"
   ],
-  "exclude": ["node_modules", "dist"]
+  "exclude": ["node_modules", "dist"],
+  "lib": ["ES2022", "DOM", "DOM.Iterable", "ESNext.Disposable"]
```
KevinOrfas: This key is at the top level of the file, next to `include` and `exclude`. `lib` is a `compilerOptions` key. TypeScript ignores unrecognised top level keys in silence, no error and no warning, so nothing told you.

```
$ npx tsc --showConfig | grep '"lib"'
(nothing)

$ # delete line 35 entirely
$ npx tsc --noEmit && echo still fine
still fine
```

And now the part that is not your fault. The story says typecheck will fail without this entry. It will not. `@types/node` v25 carries this on line 29 of its `index.d.ts`:

```ts
/// <reference lib="esnext.disposable" />
```

A `reference lib` directive makes tsc load a lib file whether you asked for it or not, so `AsyncDisposableStack` has been typed the whole time, through a package we do not own. You added the entry, typecheck went green, and there was no signal anywhere that the two facts were unrelated. The story set that up, not you.

Move it into `compilerOptions`. Not because tsc needs it today, but because `AsyncDisposableStack` is production code now and its types should not rest on a triple slash directive inside somebody else's `index.d.ts`. Keep the full list while you are there. `lib` replaces the target's defaults rather than adding to them, so trimming it to `["ESNext.Disposable"]` costs you `document` and `HTMLElement` over in `client/demo/demo.ts`. I tried it.

On the rest of question 2, worth writing in your own words in the description: `lib` only tells tsc what to assume already exists at runtime. It never emits a polyfill and never changes a line of the output JavaScript. That is precisely why raising the Node floor and adding the lib entry are two separate steps in the story rather than one. Get `lib` wrong and the build is green and the process dies on boot.

package.json
```diff
@@ -16,7 +16,7 @@
   "main": "dist/server/index.js",
   "types": "dist/server/index.d.ts",
   "engines": {
-    "node": ">=18"
+    "node": ">=24"
```
KevinOrfas, line 38: Right call, and `>=18` was fiction anyway, 18 went end of life a while back.

The paragraph the story asked for is still missing, and it matters more here than usual. There is no `.npmrc` in this repo, so `engine-strict` is off, so npm prints a warning on install and installs regardless. `>=24` documents the floor, it does not defend it. Someone on Node 22 installs EkoLite, skims the warning, and finds out at shutdown, which is the worst moment available.

The other half worth saying out loud: this drops Node 22, which is in maintenance until April 2027. That is a defensible trade for deleting a helper and I would take it. It should read as a decision we made rather than a side effect of one.

KevinOrfas (changes requested):
Read it against `main`, ran it, and went at the three questions. The swap is right. Deferred in reverse, teardown still comes out ws, then publications, then mongo. Typecheck, lint, prettier, 251 tests and CI all green.

Answer 1 is the one that mattered and you got it: teardown is the mirror of construction, which is why the stack runs backwards. `App.create` builds mongo before the socket, `close()` drops the socket before mongo, and now the registration order says so on its own rather than relying on two lists staying in step.

Three things are not done, and one of them is a behaviour change the description says did not happen. Each is in a thread, with what I ran.

- `closeAll` is still in `shared/helperFunctions.ts`. The description says it was removed.
- The `lib` entry sits at the wrong level in `tsconfig.json` and does nothing at all. That one is the story's trap rather than yours, and I explain why in the thread.
- When two closers fail, both error messages now vanish before they reach the log. That is the second half of question 3. There is a test for it in the thread on `tests/server/app.test.ts`: it passes on `main`, it fails on this branch.

Two paragraphs are still missing from the description.

The first is what `engines` actually enforces on install. That is in the thread on `package.json`.

The second is why it was safe to delete `closeAll.test.ts`, when the rule from 7.B.4 was to stop if a step seems to need a test edited. The story asks you to list every behaviour that file pinned and say where each one ended up. That is not a formality. `closeAll.test.ts` pinned two behaviours. One of them moved into `app.test.ts`. The other did not move anywhere, and it happens to be the one the swap changed. Writing that list out is the thing that would have shown you that, which is why the story asks for it in words rather than letting a green suite stand in for it.

When you reply, I would rather read what you make of the `SuppressedError` thread than see a green tick. If a piece of it does not land, say so there and we will work it out.

server/app.ts
Iam-Ebuka: About the ways to solve the SuppressedError problem, I lean towards 2 as well because it preserves all failure information and keeps App.close() honest.

But quick question, currently I am going with the option of teaching Shutdown to walk an the logging all the errors. The question is should we keep this behavior or should we collect all the errors in an array and print the heading once and then all error underneath?

tsconfig.json
Iam-Ebuka: If I got your question right, you are asking me to explain what lib is for and also why we changed the engines node version to 24. Alright, so lib is used to tell TypeScript the types that exists, it only affects type checking, it does not emit polyfills nor change the generated JavaScript. Secondly, the reason why we changed node version to 24 separately is to ensure the APIs like AsyncDisposableStack exists when application runs.

server/app.ts
KevinOrfas: Array with one heading, yes. A failed shutdown is one event, so log it as one record. Separate `console.error` calls arrive as separate lines and can interleave with the grace-timeout log firing on the same tick.

Before you build it, the bit that bites: `console.error` won't print the two messages for you. A `SuppressedError` shows its own message and stack, but `.error` and `.suppressed` are non-enumerable, so `socket close failed` and `mongo close failed` never appear. You have to pull them out yourself.

So put the walk in a small pure helper, `flattenSuppressed(err): unknown[]`, that follows `.suppressed` down to the leaf and returns the errors oldest-first. Shutdown logs that list under one heading. The helper is where the test goes (see the app.test.ts thread), because the helper is pure and `console.error` isn't something we can assert without a spy.

tests/server/app.test.ts
KevinOrfas: Under option 2, `App.close` keeps rejecting with the raw `SuppressedError`, and the "name every failure" behaviour becomes the flatten. That's the pure part, so that's what to pin, no seam and no spy:

```ts
import { flattenSuppressed } from '../../shared/helperFunctions.ts';

// AsyncDisposableStack hides each failure on a SuppressedError's non-enumerable
// .error / .suppressed. flattenSuppressed pulls them back out so Shutdown can log
// every closer that broke, not just the last one.
describe('flattenSuppressed', () => {
  it('returns every error in a suppressed chain, oldest first', () => {
    const socket = new Error('socket close failed');
    const mongo = new Error('mongo close failed');
    const chain = new SuppressedError(mongo, socket);

    expect(flattenSuppressed(chain).map((e) => (e as Error).message)).toEqual([
      'socket close failed',
      'mongo close failed',
    ]);
  });

  it('returns a lone plain error unchanged', () => {
    const only = new Error('socket close failed');
    expect(flattenSuppressed(only)).toEqual([only]);
  });
});
```

Red on this branch because `flattenSuppressed` doesn't exist yet. It's the natural replacement for `closeAll` in `shared/helperFunctions.ts`: that helper leaves because it named the first failure, this one arrives so Shutdown can name all of them.

server/app.ts
```diff
@@ -83,9 +83,11 @@ export class App {
   }
 
   static createNull(nullConfig: NullConfig = {}): App {
-    const mongo = nullConfig.findResponses
-      ? MongoWrapper.createNull({ find: nullConfig.findResponses })
-      : MongoWrapper.createNull();
+    const mongo =
+      nullConfig.mongo ??
```
KevinOrfas, line 91: Passing both `mongo` and `findResponses` silently drops `findResponses`: `mongo` wins the `??` and the find branch never runs. Nothing calls it that way today, but the next person who does gets an empty nulled Mongo and no error saying why. Either apply `findResponses` to the injected mongo, or throw on the ambiguous combo so it fails loudly.

If you go the throw route, this pins it (red today, and it encodes that choice):

```ts
it('rejects a config that sets both mongo and findResponses', () => {
  const mongo = MongoWrapper.createNull();
  expect(() => App.createNull({ mongo, findResponses: [[]] })).toThrow(/both|ambiguous/i);
});
```

KevinOrfas (changes requested):
Second pass, on `1db13c9`. CI is green, and you beat me to the lint errors. I was still reading the old head when you pushed the fix.

The heart of it is done and it is right. `flattenSuppressed` digs the real errors back out of the `SuppressedError`, and `Shutdown` logs all of them, so an operator can see what broke. I pushed a three failure shutdown through your recursion to check it holds at depth, and it does.

Rather than list what I think, here is a file. Drop it in and run it:

```
npx vitest run tests/server/closeContract.review.test.ts
```

```ts
import { describe, expect, it } from 'vitest';
import { App } from '../../server/app.ts';
import { MongoWrapper } from '../../server/infrastructure/mongo.ts';
import { WebSocketWrapper } from '../../server/infrastructure/websocket.ts';
import { flattenSuppressed } from '../../shared/helperFunctions.ts';

describe('what App.close() rejects with', () => {
  // 1. The test app.test.ts already carries, by name. The name says 'the first
  //    failure'. This is the assertion that name describes.
  it('rejects with the first failure, so Shutdown can name it', async () => {
    const mongo = MongoWrapper.createNull({ close: [new Error('mongo close failed')] });
    const ws = WebSocketWrapper.createNull({ close: [new Error('socket close failed')] });
    const app = App.createNull({ mongo, ws });

    await expect(app.close()).rejects.toThrow('socket close failed');
  });

  // 2. What it actually does now, which is what we agreed it should do. Pins the
  //    type, not the message: 'An error was suppressed during disposal' is V8's
  //    wording, not ours, and a Node upgrade can change it.
  it('rejects with a SuppressedError carrying every failure, oldest first', async () => {
    const mongo = MongoWrapper.createNull({ close: [new Error('mongo close failed')] });
    const ws = WebSocketWrapper.createNull({ close: [new Error('socket close failed')] });
    const app = App.createNull({ mongo, ws });

    const err: unknown = await app.close().catch((e: unknown) => e);

    expect(err).toBeInstanceOf(SuppressedError);
    expect(flattenSuppressed(err).map((e) => (e as Error).message)).toEqual([
      'socket close failed',
      'mongo close failed',
    ]);
  });

  // 3. App.close() defers three closers. Every test we have only ever fails two.
  //    Two is the case where a wrong recursion still looks right. This is three.
  it('returns every error in a three deep chain, oldest first', async () => {
    const stack = new AsyncDisposableStack();
    stack.defer(() => Promise.reject(new Error('mongo close failed')));
    stack.defer(() => Promise.reject(new Error('publications stopAll failed')));
    stack.defer(() => Promise.reject(new Error('socket close failed')));

    const err: unknown = await stack.disposeAsync().catch((e: unknown) => e);

    expect(flattenSuppressed(err).map((e) => (e as Error).message)).toEqual([
      'socket close failed',
      'publications stopAll failed',
      'mongo close failed',
    ]);
  });
});
```

```
Tests  1 failed | 2 passed (3)

× rejects with the first failure, so Shutdown can name it
  Expected: "socket close failed"
  Received: "An error was suppressed during disposal"
```

The one that fails is the one whose name you already have in `app.test.ts`. `App.close()` does not reject with the first failure any more, and it should not, because that was the decision. So it is the sentence on top that still belongs to `closeAll`, not the code underneath it. Test 2 is what I would put in its place. Test 3 is the gap: keep it, it is cheap and nothing covers that shape today.

Then four leftovers, each one a place the code moved and something around it did not.

- `shutdown.ts:44`, `logShutdowError` is missing its n.
- `shutdown.ts:46`, that `if` asks a question `flattenSuppressed` has already answered. Your own `returns a lone plain error unchanged` is the proof it is covered. `console.error('shutdown failed:', ...flattenSuppressed(err))` and the branch goes.
- `websocket.ts:9` and `:77`. `CLOSE_EVENT` is emitted, but `emitter` is private and `trackClose()` is gone, so nothing in the language can hear it. I deleted both lines and the suite stayed at 257 green, which is the proof there was no listener.
- `flattenSuppressed.test.ts` is in `tests/server` but tests `shared/helperFunctions.ts`. `tests/shared/helperFunctions.test.ts` is next door already.

And the description, which still says the behaviour is identical. It is not, and that is the interesting bit rather than a formality. `App` is exported from `server/index.ts`, so `App.close()` is published API and what it throws has changed. While you are there: `"node": ">=24"` only warns on install, there is no `.npmrc` in the repo, so someone on Node 22 installs, boots fine, and meets `AsyncDisposableStack is not defined` at shutdown, which is the worst moment available.

Which is my question, and I would rather read your answer than see a green tick. When we picked option 2 I was thinking about our logs. I was not thinking about somebody else's `catch` block. Is `SuppressedError` the right thing to hand a consumer, or should `App.close()` reject with something plainer and let `Shutdown` flatten it where we own both ends? I lean towards what you built. You have been in this code a fortnight and I have not.

Once you have read the file, delete it: test 2 replaces the one in `app.test.ts`, test 3 goes to `tests/shared` with the others.

tests/helpers/flattenSuppressed.test.ts
```diff
@@ -0,0 +1,23 @@
+import { describe, expect, it } from 'vitest';
+import { flattenSuppressed } from '../../shared/helperFunctions.ts';
+
+// AsyncDisposableStack hides each failure on a SuppressedError's non-enumerable
+// .error / .suppressed. flattenSuppressed pulls them back out so Shutdown can log
+// every closer that broke, not just the last one.
+describe('flattenSuppressed', () => {
```
KevinOrfas: This one tests `shared/helperFunctions.ts`. `tests/helpers` is where the test infrastructure lives, things like `freePort.ts` and `serverProcess.ts`, so it is the tools the suite uses rather than the code the suite checks. Moving it out of `tests/server` was right, it has just landed one door along.

`tests/shared/helperFunctions.test.ts` already exists and already covers `hasMongoOperator` from the same file. Move the `describe` block in there and delete this one. The three deep chain test from the other thread can go in beside it, since it is really testing `flattenSuppressed` rather than `App`.

KevinOrfas (changes requested):
Third pass, on `8c9b0d5`. You have been moving faster than I have been reading, and most of Monday is already in: `closeAll` is out of `helperFunctions.ts`, `lib` is inside `compilerOptions` with the full list kept, the unused `trackClose` and `CLOSE_EVENT` are gone, `MongoWrapper` has the close seam the pin needed, and `createNull` now throws on the ambiguous config with a test on it. I ran `app.test.ts`, `flattenSuppressed.test.ts` and `shutdown.test.ts` at head and got 19 green.

Four threads left. Start with the `package.json` one, because the `.npmrc` does less than it looks like it does and it is better to know that before you build anything on top of it.

On the description. It no longer says the behaviour is identical, and answer 3 now explains the difference, so that is sorted. The bit still missing is what it means for other people. `App` is exported from `server/index.ts`, which is the main entry of the published package, so `App.close()` is public API and what it throws has changed here. Anyone with a `catch` around it sees something different after upgrading. Worth a couple of lines in the description, since that is where people look. Same for Node: this drops Node 22, which is supported until April 2027. I think that is a good trade for what we get, but let us say so on purpose rather than leave it as a bullet.

Two small ones:

* `shutdown.ts`, `logShutdownError` finishes with a bare `return;`. It had a job when there was a branch above it, and now it does not.
* `app.ts`, `createNull` has a ternary inside a ternary now. `nullConfig.mongo ?? (nullConfig.findResponses ? ... : ...)` is the same thing and reads in one go.

On the threads themselves, either way works. If one of them makes sense to you and you reckon it is done, close it and leave a line saying what you took from it. If you would rather I had another look before it closes, leave it open and nudge me, and I will check and close it from my side. And if one is still murky, say that too and we will keep it open and talk it through. The line matters to me more than who clicks the button. Green tests tell me the code moved, they do not tell me we see it the same way.

tests/server/app.test.ts
KevinOrfas: Two things have got tangled here.

```ts
it('rejects with the first failure, so Shutdown can name it', async () => {
  ...
  await expect(closePromise).rejects.toThrow('An error was suppressed during disposal');
```

The name came from the `closeAll` days. `App.close()` does not reject with the first failure any more, and we agreed it should not, so the title now describes the behaviour this PR removes. Someone reading the suite to learn what `App.close()` does would come away with the wrong idea.

The string is a separate thing. `An error was suppressed during disposal` is not in the spec. The spec says a `SuppressedError` has a message, it does not say what the message is, so that wording comes from V8 and a Node upgrade could change it. If that happens the test goes red even though nothing we promised has broken.

Better to pin the type, and pin the contents with the helper you wrote:

```ts
it('rejects with a SuppressedError carrying every failure, oldest first', async () => {
  const mongo = MongoWrapper.createNull({ close: [new Error('mongo close failed')] });
  const ws = WebSocketWrapper.createNull({ close: [new Error('socket close failed')] });
  const app = App.createNull({ mongo, ws });

  const err: unknown = await app.close().catch((e: unknown) => e);

  expect(err).toBeInstanceOf(SuppressedError);
  expect(flattenSuppressed(err).map((e) => (e as Error).message)).toEqual([
    'socket close failed',
    'mongo close failed',
  ]);
});
```

One more reason to move to that shape. In your version the expectations sit inside `await closePromise.catch(err => { ... })`, and a `.catch` block only runs if the promise rejects. If `App.close()` ever stops rejecting, those expectations are skipped and the test still goes green. The `rejects.toThrow` line above happens to cover you today, so it is not a bug right now, but it is a shape worth avoiding.

The other half is the three deep chain, which is still missing. `App.close()` defers three closers, and every test we have only fails two. Two is the size where a wrong version still passes: if you skipped the recursion entirely and just wrote `[err.suppressed, err.error]`, the two failure test goes green. At three, `err.suppressed` is itself a `SuppressedError`, so that flat version hands back a nested error where a message should be. Your recursion is right. The suite just has no way of showing that yet. Test 3 from the file I sent is the one that does, and it belongs with the other `flattenSuppressed` tests, see the thread on that file.

package.json
KevinOrfas: The `.npmrc` is a good instinct, though it does not quite do what the commit message says.

`engine-strict` is read from the config of whoever is running the install. npm does not read a dependency's `.npmrc`, and it does not publish one either, it is on npm's always ignored list along with `.git` and `node_modules`. I checked with `npm pack --dry-run` and it is not in the tarball.

So what the file does is make `npm install` inside this repo fail on Node 22 rather than warn. That is genuinely useful and I would keep it, it protects us. It just does not reach the person running `npm install ekolite` on Node 22. They still get a warning they will probably skim past, then a clean install, a server that boots fine, and `AsyncDisposableStack is not defined` at shutdown, which is a rough place to find out.

That is what I meant by `>=24` documenting the floor without defending it. The description is the thing that actually helps here, because it is the one place a consumer will read. A couple of lines: the floor is Node 24, npm will not stop you going below it, and here is what breaks if you do.

server/app.ts
KevinOrfas: > I think exposing the actual error can help the user to understand properly what the issue is incase they need more context to an error they are facing.

Agreed, and it is where I land too, so let us keep what you built. There is one loose end worth pulling, though, and it only shows up from the consumer's side.

Say someone catches our `SuppressedError` and does the obvious thing:

```js
try {
  await app.close();
} catch (err) {
  console.error(err);
}
```

They get the message and a stack and nothing else, because `.error` and `.suppressed` are non enumerable, so both real failures stay hidden on the object. That is exactly the trap we hit in `Shutdown`, and `flattenSuppressed` is how we got out of it.

The snag is that `flattenSuppressed` lives in `shared/helperFunctions.ts`, and the `./shared` export of the package points at `shared/protocol.ts`, so a consumer cannot reach it. At the moment we throw an error they have no easy way to read.

Exporting `flattenSuppressed` from `server/index.ts` alongside `App` closes that. The thing that throws and the thing that reads it then travel together, and your answer holds all the way out to the consumer.

